PR #19, opened 2022-05-11
Update column + temporal filtering

erezsh:


data_diff/diff_tables.py
```diff
@@ -23,23 +23,36 @@ class TableSegment:
...
     _checksum: int = None
 
-    def _make_range_pred(self):
-        if self.start is not None:
-            yield Compare("<=", str(self.start), self.key_column)
-        if self.end is not None:
-            yield Compare("<", self.key_column, str(self.end))
+    def __post_init__(self):
+        if not self.update_column and (self.min_time or self.max_time):
+            raise ValueError("Error: min_time/max_time feature requires to specify 'update_column'")
```
sirupsen, line 38: 👍🏻

data_diff/diff_tables.py
```diff
@@ -23,23 +23,36 @@ class TableSegment:
...
+        if self.start_key is not None:
+            yield Compare("<=", str(self.start_key), self.key_column)
+        if self.end_key is not None:
+            yield Compare("<", self.key_column, str(self.end_key))
+
+    def _make_update_range(self):
+        if self.min_time is not None:
+            yield Compare("<=", Time(self.min_time), self.update_column)
+        if self.max_time is not None:
+            yield Compare("<", self.update_column, Time(self.max_time))
```
sirupsen, line 50: A Python question: Why is `yield` preferred here to just returning a `list`? Because it's annoying to merge the lists I'm assuming, versus adding generators together and then materializing them to a list like you're doing below? Kinda cool pattern.

data_diff/diff_tables.py
```diff
@@ -23,23 +23,36 @@ class TableSegment:
...
-            yield Compare("<", self.key_column, str(self.end))
+    def __post_init__(self):
+        if not self.update_column and (self.min_time or self.max_time):
+            raise ValueError("Error: min_time/max_time feature requires to specify 'update_column'")
+
+    def _make_key_range(self):
+        if self.start_key is not None:
+            yield Compare("<=", str(self.start_key), self.key_column)
+        if self.end_key is not None:
+            yield Compare("<", self.key_column, str(self.end_key))
```
sirupsen, line 44: I think the key version will also benefit from a CLI arg 👍🏻 Thanks for changing its name and making it symmetric with time

Lots of people will probably instead of temporal just get the id of the last records when they did the last insertion, then do the check based on that

tests/test_diff_tables.py
```diff
@@ -2,36 +2,110 @@
...
+                comment: string
+            }
+            commit()
+
+            func add(date, comment) {
+                new a(date, comment)
+            }
+        """)
+        self.now = now = arrow.get(self.preql.now())
+        self.preql.add(now.shift(days=-50), "50 days ago")
```
sirupsen, line 36: When I try to run your tests, this line errors (and presumably the others will too):

CleanShot 2022-05-11 at 15 45 02@2x

Would love if `preql` gave some more useful output here 😅

tests/test_diff_tables.py
```diff
@@ -2,36 +2,110 @@
...
+
+    def test_basic(self):
+        differ = TableDiffer(10, 100)
+        a = TableSegment(self.connection, ('a', ), 'id', 'datetime')
+        b = TableSegment(self.connection, ('b', ), 'id', 'datetime')
+        assert a.count == 6
+        assert b.count == 5
+
+        assert not list(differ.diff_tables(a, a))
+        self.assertEqual( len( list(differ.diff_tables(a, b)) ), 1 )
```
sirupsen, line 63: Remember to run `black(1)`, which'll probably complain about the spaces here

tests/test_diff_tables.py
```diff
@@ -2,36 +2,110 @@
...
+
+    def test_offset(self):
```
sirupsen, line 65: What about a test where the diff actually passes when you pass a range?

sirupsen (commented):
Couple of comments, but basically looks good to me—you can merge this without me looking at it again if you prefer

data_diff/diff_tables.py
erezsh: Yes, it's just a cleaner syntax, saving me from creating a list, appending to it, and returning it.

data_diff/diff_tables.py
erezsh: We already have `-k`. Do you mean something different?

tests/test_diff_tables.py
erezsh: This doesn't look related to `preql` in any way. What is Python.Framework? Never heard of it.

tests/test_diff_tables.py
erezsh: Good point, thanks

tests/test_diff_tables.py
erezsh: That's what I do at the end of `test_offset`, unless you mean something else

PR #20, opened 2022-05-11
readme: recommend just 1m rows for development

sirupsen:
It seems a bit more reasonable to just import 1M rows for development to not make it take 10min or so to get started, even longer on other databases. 1M is plenty for getting started

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,40 +137,45 @@ directly with `poetry` rather than go through the package.
 brew install mysql postgresql # MacOS dependencies for C bindings
 poetry install
 ```
-
-**2. Download CSV of Testing Data**
-
-```shell-session
-wget https://files.grouplens.org/datasets/movielens/ml-25m.zip
-unzip ml-25m.zip -d dev/
-```
-
-**3. Start Databases**
+**2. Start Databases**
 
 ```shell-session
 docker-compose up -d mysql postgres
 ```
 
-**4. Run Unit Tests**
+**3. Run Unit Tests**
 
 ```shell-session
 poetry run python3 -m unittest
 ```
 
-**5. Seed the Database(s)**
+**4. Seed the Database(s)**
 
-If you're just testing, we recommend just setting up one database (e.g.
-Postgres) to avoid incurring the long setup time repeatedly.
+First, download the CSVs of seeding data:
 
 ```shell-session
-preql -f dev/prepare_db.pql postgres://postgres:Password1@127.0.0.1:5432/postgres
+wget https://files.grouplens.org/datasets/movielens/ml-1m.zip
+unzip ml-1m.zip -d dev/ && mv dev/ml-1m dev/ml
+
+# For a larger data-set (but takes longer to import):
+# - https://files.grouplens.org/datasets/movielens/ml-25m.zip
+
+```
+
+Now you can insert it into the testing database(s):
+
+```shell-session
+# It's optional to seed more than one to run data-diff(1) against.
 preql -f dev/prepare_db.pql mysql://mysql:Password1@127.0.0.1:3306/mysql
-preql -f dev/prepare_db.pql snowflake://<uri>
-preql -f dev/prepare_db.pql mssql://<uri>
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -136,40 +136,48 @@ directly with `poetry` rather than go through the package.
...
 
-**4. Run Unit Tests**
+**4. Seed the Database(s)**
+
+First, download the CSVs of seeding data:
 
 ```shell-session
-poetry run python3 -m unittest
-```
+wget https://files.grouplens.org/datasets/movielens/ml-25m.zip
```
erezsh, line 156: Link to 1m rows ;)

https://files.grouplens.org/datasets/movielens/ml-1m.zip

There's also 100k if you want it even faster. (see https://files.grouplens.org/datasets/movielens )

README.md
```diff
@@ -136,40 +136,48 @@ directly with `poetry` rather than go through the package.
...
-If you're just testing, we recommend just setting up one database (e.g.
-Postgres) to avoid incurring the long setup time repeatedly.
+Now you can insert it into the testing database(s):
 
 ```shell-session
 preql -f dev/prepare_db.pql postgres://postgres:Password1@127.0.0.1:5432/postgres
+# You can compare in just one if you like, you don't need to seed them all.
 preql -f dev/prepare_db.pql mysql://mysql:Password1@127.0.0.1:3306/mysql
+
+# These typically would run remotely
```
erezsh, line 174: Categorize them by classical / cloud dbs . then it's obvious

README.md
```diff
@@ -136,40 +136,48 @@ directly with `poetry` rather than go through the package.
...
-**5. Seed the Database(s)**
+```
 
-If you're just testing, we recommend just setting up one database (e.g.
-Postgres) to avoid incurring the long setup time repeatedly.
+Now you can insert it into the testing database(s):
 
 ```shell-session
 preql -f dev/prepare_db.pql postgres://postgres:Password1@127.0.0.1:5432/postgres
+# You can compare in just one if you like, you don't need to seed them all.
```
erezsh, line 171: This comment here is out of place. But we can write at the top "These are optional, based on what dbs you have. Just one of them is enough to run everything."

Also they *need* mysql to run tests. Unless we switch the default to postgres.

README.md
```diff
@@ -136,40 +136,48 @@ directly with `poetry` rather than go through the package.
...
 preql -f dev/prepare_db.pql mysql://mysql:Password1@127.0.0.1:3306/mysql
+
+# These typically would run remotely
 preql -f dev/prepare_db.psq snowflake://<uri>
 preql -f dev/prepare_db.psq mssql://<uri>
 preql -f dev/prepare_db_bigquery.pql bigquery:///<project> # Bigquery has its own
 ```
 
-**6. Run data-diff against seeded database**
+**5. Run `data-diff` against seeded database**
```
erezsh, line 180: I actually think it's better to leave it as text, rather than code, as it's a name. Maybe put it in bold instead?

README.md
```diff
@@ -180,7 +188,10 @@ Diff-Split: +250156  -0
 ```
 
 # How to publish to PyPI
```
erezsh, line 190: Why do we have instructions on publishing to PyPI? We are the only ones who will do it. (and probably 1 specific person who's in charge of it)

erezsh (changes requested):

README.md
erezsh: I think we should remove it. It will confuse users, if nothing else.

It's fine if we create PUBLISH.md if we just want to be perfectly open, I just don't think it belong in the README.

README.md
sirupsen: Ya, but you don't _need_ to populate MySQL to run tests (and we should keep it that way)

sirupsen:
@erezsh tried addressing your comments, good to go? ✅

README.md
erezsh: Fair point

erezsh (approved):

PR #21, opened 2022-05-12
Added --min-age and --max-age to CLI, time parsing

erezsh:


data_diff/parse_time.py
```diff
@@ -0,0 +1,61 @@
+import re
```
sirupsen, line 1: Can you create some unit tests for this? 🙏🏻

data_diff/parse_time.py
```diff
@@ -0,0 +1,61 @@
...
+    return SequenceMatcher(None, a, b).ratio()
+
+
+def parse_time_atom(count, unit):
+    count = int(count)
+    try:
+        unit = TIME_UNITS[unit]
+    except KeyError:
+        most_similar = max(TIME_UNITS, key=lambda k: string_similarity(k, unit))
+        raise ValueError(f"'{unit}' is not a recognized time unit. Did you mean '{most_similar}'?")
```
sirupsen, line 39: ❤️ this attention to detail!

sirupsen:
`ValueError: Error: min_time/max_time feature requires to specify 'update_column'`

Can you include in this error to add `--update-column` if using the CLI, if you're using exceptions for user errors? 👍🏻

data_diff/parse_time.py
```diff
@@ -0,0 +1,61 @@
...
+
+    if unit in EXTRAPOLATED:
+        mul, unit = EXTRAPOLATED[unit]
+        count *= mul
+    return count, unit
+
+
+def parse_time_delta(t: str):
```
sirupsen, line 47: What if I pass `1h30min`? I think it'll just interpret it as `1h`, even though that's not what I meant. Probably worth erroring in this case?

sirupsen: What about floats? E.g. `1.5h` isn't crazy

We should probably support either floats or using multiple deltas like `1h30min`.

data_diff/parse_time.py
```diff
@@ -0,0 +1,61 @@
+import re
+from datetime import datetime, timedelta
+from difflib import SequenceMatcher
```
sirupsen, line 3: :)

sirupsen (changes requested):

data_diff/parse_time.py
sirupsen: Probably worth printing all units too

data_diff/parse_time.py
erezsh: We already support `1h30min` ;)

data_diff/parse_time.py
erezsh: Not sure what to do with a smiley :)

sirupsen: just smile!

data_diff/parse_time.py
sirupsen: I missed the while! 👍🏻 👍🏻

erezsh:
@sirupsen I think I addressed all your concerns

sirupsen (approved):

sirupsen:
@erezsh forgot to say: can you add these to the README too?

erezsh:
Done

PR #23, opened 2022-05-12
readme: add debian based deps

sirupsen:
@cfernhout @erezsh I installed this on my linux box to make sure the instructions work, and this is what I needed to install 👍🏻


erezsh:
LGTM

PR #24, opened 2022-05-13
Better prepare_db to use latest preql features

erezsh:
A step in the right direction for better integration with preql

### DIFF
--- a/dev/prepare_db.pql
+++ b/dev/prepare_db.pql
@@ -8,6 +8,12 @@ func drop_table(t) {
     run_sql("DROP TABLE IF EXISTS " + t)
 }
 
+func create_indices(tbl) {
+    tbl.add_index("id", true)
+    tbl.add_index("timestamp")
+    tbl.add_index(["id", "timestamp"])
+}
+
 // Cleanup
 func cleanup() {
     drop_table("rating")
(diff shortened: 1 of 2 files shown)


erezsh:
I think next step will be to merge the separate bigquery file into prepare_db

dev/prepare_db.pql
```diff
@@ -4,41 +4,73 @@ func run_sql(code) {
...
 }
 
+func drop_table(t) {
+    run_sql("DROP TABLE IF EXISTS " + t)
+}
+
+func create_indices(tbl) {
+    tbl.add_index("id", true)
+    tbl.add_index("timestamp")
+    tbl.add_index(["id", "timestamp"])
```
sirupsen, line 14: Ideally these would be created as part of the table definition instead of built afterward. It's nicer if that's defined together

dev/prepare_db.pql
```diff
@@ -4,41 +4,73 @@ func run_sql(code) {
...
-    run_sql("DROP TABLE IF EXISTS rating_del1p")
-    run_sql("DROP TABLE IF EXISTS rating_update50p")
+    drop_table("rating")
+    drop_table("tmp_rating")
+    drop_table("rating_del1")
+    drop_table("rating_update1")
+    drop_table("rating_update001p")
+    drop_table("rating_update1p")
+    drop_table("rating_del1p")
+    drop_table("rating_update50p")
```
sirupsen, line 26: Should `.drop` on the table definition be a `preql` construct?

dev/prepare_db.pql
```diff
@@ -4,41 +4,73 @@ func run_sql(code) {
...
 
 cleanup()
 
 // Import CSV
-if (db_type == "snowflake") {
-    print "Uploading ratings CSV"
-
-    run_sql("RM @~/ratings.csv.gz")
-    run_sql("PUT file://dev/ratings.csv @~")
+if (db_type == "snowflake" or db_type == "redshift") {
```
sirupsen, line 33: What could we do to remove the Cloud databases as special-cases?

sirupsen (commented):
Did you get this against all of them? 

How can we make it easy for others (inside the Datafold org), to do the same?

dev/prepare_db.pql
erezsh: Generally I agree, but I'm not sure what's the best way to do it, syntax-wise.

Especially on (id, timestamp).

Also, does it mean that indexes are copied when you copy a table?

This just seemed like the simplest, most robust solution.

dev/prepare_db.pql
erezsh: There is already `remove_table_if_exists`, which does the same thing. But it doesn't print out the SQL.

dev/prepare_db.pql
erezsh: I can do something like `db.is_cloud()`, but that would only solve this specific `if`.

I don't see how we can get rid of db-specific code.

What we can do is have an separate module for each db, implementing `import_sample_csv()`, and have the main module call it. So at least the db-specific stuff will be separated.

erezsh:
> Did you get this against all of them?
> 
> How can we make it easy for others (inside the Datafold org), to do the same?

Not sure I understand the question.. You just need to set up the databases and run it.

PR #25, opened 2022-05-13
Chiel dat 3282 set up ci

cfernhout:
Basic setup for automatically testing unit tests on two versions of Python (3.7 & 3.10). Interestingly, some unit tests fail for Python 3.7.


sirupsen (commented):
@cfernhout sorry I was super unclear on Slack. Code here LGTM, I think my question for whether we should do more checks is basically: How much longer do I have to wait? How much more flaky is it likely to be?

I would love CI to pass in 2min for everything. If it takes much longer, I think we can peel some of these versions/OS combinations off, at your discretion

cfernhout:
@sirupsen For now having a matrix of 4 iterations will run almost always in parallel. Only if the number of jobs explode (approx > 10) then you will get to the point that there is a chance that you need to wait on machines to become available. For now, everything runs quickly.

PR #26, opened 2022-05-13
Chiel dat 3282 set up ci multiple db

cfernhout:
Now added testing for also Postgres with the unit test as is. Postgres unit tests are mostly failing for now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,8 +23,8 @@ jobs:
           - "3.9"
           - "3.10"
         db-conn: [
-          {db: mysql, conn: mysql://mysql:Password1@localhost/mysql},
-          {db: postgres, conn: postgres://postgres:Password1@localhost/postgres}
+          {db: mysql, conn: "mysql://mysql:Password1@localhost/mysql"},
+          {db: postgres, conn: "postgres://postgres:Password1@localhost/postgres"}
         ]
 
     name: Check Python ${{ matrix.python-version }} on ${{ matrix.os }} for ${{ matrix.db-conn.db }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,7 +38,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
       
       - name: Build the stack
-        run: docker-compose up -d mysql
+        run: docker-compose up -d ${{ matrix.db-conn.db }}
 
       - name: Install Poetry
         run: pip install poetry
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,8 +22,12 @@ jobs:
           - "3.8"
           - "3.9"
           - "3.10"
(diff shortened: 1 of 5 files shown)


tests/common.py
```diff
@@ -1,11 +1,12 @@
 import hashlib
+import os
 
 from data_diff.database import CHECKSUM_HEXDIGITS, MD5_HEXDIGITS
 import logging
 
 logging.basicConfig(level=logging.WARN)
 
-TEST_MYSQL_CONN_STRING = "mysql://mysql:Password1@localhost/mysql"
+TEST_CONN_STRING = os.environ.get("TEST_CONN_STRING", "mysql://mysql:Password1@localhost/mysql")
```
erezsh, line 9: It's probably better to remove the default and throw an error message Because the error "bad username" or "no such database" can be confusing, and no one will know to look here for it.

.github/workflows/ci.yml
```diff
@@ -22,8 +22,12 @@ jobs:
           - "3.8"
           - "3.9"
           - "3.10"
+        db-conn: [
+          {db: mysql, conn: "mysql://mysql:Password1@localhost/mysql"},
+          {db: postgres, conn: "postgres://postgres:Password1@localhost/postgres"}
+        ]
```
erezsh, line 28: I like how it's progressing!

sirupsen:
Hmmm... it's interesting to write the suite generically, and then have CI run against all of them. What I don't love about it though is that it doesn't run them all locally, but that also doesn't seem completely realistic... hmmm

erezsh:
If you're referring to the cloud databases, maybe we can randomize the dataset name, so there won't be any collisions? (unless that's not the issue)

tests/common.py
cfernhout: Sounds reasonable. Will change that.

cfernhout: My idea initially to use this as a default is that nothing would change. FYI ;)

cfernhout:
@sirupsen If I understand you correctly, the thing that you don't love is that you need to fire multiple commands with multiple environment variables to run them all locally? I think we could use a make file, such that it is still only one command. Like `make test` which runs basically them all.

cfernhout:
@erezsh The tests are not referring to cloud databases. They could be empty databases for all that matter. I don't think we want to make our unit tests dependent on external databases. If we would write other types of tests, like integration tests or performance tests, then using external cloud databases would be reasonable.

erezsh:
@sirupsen It makes sense to me to have "short tests" that run for each commit, and "full tests" that run before a release, which can include the cloud support.

cfernhout:
@erezsh Can you be more specific what you want to test with "full tests"? Do you mean, performance, integration (with other services), more edge cases, just more test coverage, etc?

erezsh:
It's always good to have more of everything, but I was talking about basic coverage - to see the tool runs on every database with every feature enabled.

PR #30, opened 2022-05-18
Threading Support

erezsh:


erezsh:
Wrong implementation..

erezsh:
This seems to be okay now, as a first implementation. Main disadvantage of this approach is that we may end up opening a lot of threads. However, it's not necessarily that bad. We can also add measures to curb the thread count and keep it within a reasonable limit.

An alternative would be to use futures/async, instead of regular threads, which will reduce the overhead, but probably complicate the code.

data_diff/database.py
```diff
@@ -52,18 +54,18 @@ def _one(seq):
     return x
 
 
+def _query_conn(conn, sql_code: str) -> list:
+    c = conn.cursor()
+    c.execute(sql_code)
+    return c.fetchall()
```
sirupsen, line 60: Why is this not on the `ThreadedDatabase`? 👂🏻

data_diff/database.py
```diff
@@ -110,6 +112,38 @@ def md5_to_int(self, s: str) -> str:
...
+        assert not hasattr(self.thread_local, "conn")
+        self.thread_local.conn = self.create_connection()
+
+    def _query(self, sql_code: str):
+        r = self._queue.submit(self._query_in_worker, sql_code)
+        return r.result()
+
+    def _query_in_worker(self, sql_code: str):
+        "This method runs in a worker thread"
+        return _query_conn(self.thread_local.conn, sql_code)
```
sirupsen, line 141: I'm not too familiar with error handling in threads in Python–if an exception is raised in the thread pool, is it propagated to kill the main thread? 👂🏻 I think it should. 

I ask because e.g. in Ruby/Go, that's not the case

data_diff/diff_tables.py
```diff
@@ -148,6 +150,13 @@ def diff_sets(a: set, b: set) -> iter:
 
 DiffResult = iter  # Iterator[Tuple[Literal["+", "-"], tuple]]
 
+def thread_map(func, iter):
+    task_pool = ThreadPoolExecutor()
+    return task_pool.map(func, iter)
+
+def precalc_attr(attr, iter):
```
sirupsen, line 157: `precalc_` doesn't imply to me it's threaded, is this naming a Python pattern I am not aware of? Could we call it `threaded_map_attr` or something?

data_diff/database.py
```diff
@@ -110,6 +112,38 @@ def md5_to_int(self, s: str) -> str:
...
+        ...
+
+
+class ThreadedDatabase(Database):
+    """Access the database through singleton threads.
+
+    Used for database connectors that do not support sharing their connection between different threads.
+    """
+
+    def __init__(self, thread_count=1):
```
sirupsen, line 127: In a follow-up you'll allow specifying the concurrency in the CLI? 👍🏻

data_diff/database.py
sirupsen: Ahh... it's because it's not threaded. If this is the normal pattern rather than instantiating a `SingleThreaded` or just with 1 thread, all good

data_diff/diff_tables.py
```diff
@@ -211,14 +223,18 @@ def _diff_tables(self, table1, table2, level=0):
...
-            if t1.checksum != t2.checksum:
-                # Apply recursively
-                yield from self._diff_tables(t1, t2, level + 1)
+        # Recursively compare each pair of corresponding segments between table1 and table2
+        diff_iters = [
+            self._diff_tables(t1, t2, level + 1, i + 1, len(segmented1))
+            for i, (t1, t2) in enumerate(safezip(segmented1, segmented2))
+        ]
+
+        for res in thread_map(list, diff_iters):
```
sirupsen, line 239: I wonder whether it's worth it to wait with printing the verbose output until this point too... so you can sort it by segment. I see pros and cons with each...

sirupsen: CleanShot 2022-05-19 at 10 38 47@2x

data_diff/database.py
sirupsen: Even with this default though, it does seem to use concurrency?

```
[10:39:11] INFO - . Diffing segment 1/12 of size 90908 and 90908
[10:39:11] INFO - . Diffing segment 8/12 of size 90909 and 90909
[10:39:11] INFO - . Diffing segment 9/12 of size 90909 and 90909
[10:39:11] INFO - . Diffing segment 10/12 of size 90909 and 90909
[10:39:11] INFO - . Diffing segment 11/12 of size 90909 and 90909
[10:39:11] INFO - . Diffing segment 12/12 of size 2 and 2
[10:39:11] INFO - . Diffing segment 7/12 of size 90909 and 90909
[10:39:11] INFO - . Diffing segment 5/12 of size 90909 and 90909
[10:39:11] INFO - . Diffing segment 6/12 of size 90909 and 90908
[10:39:11] INFO - . Diffing segment 4/12 of size 90909 and 90909
[10:39:11] INFO - . Diffing segment 2/12 of size 90909 and 90909
[10:39:11] INFO - . Diffing segment 3/12 of size 90909 and 90909
```

Why isn't it basically single-threaded with `thread_count=1`?

sirupsen (commented):
Overall no major qualms here 🎉 
You've clearly gone through a few iterations here to make it as simple as possible! 👏🏻

data_diff/database.py
erezsh: That's how it's supposed to work. I will verify.

data_diff/database.py
erezsh: Yes, the CLI will pass an argument for that, once we figure out the interface.

The database access is single-threaded (when =1), but the algorithm itself is still threaded. That's necessary for taking advantage of thread_count>1. I could limit the algorithm to 1 thread too in this case, but I'm not sure that it matters.

data_diff/diff_tables.py
sirupsen: I like threading in it if you don't mind

data_diff/database.py
sirupsen: I think it matters because in that case I'd expect to see the output in order if threads=1

data_diff/diff_tables.py
sirupsen: Yeah, I agree... I don't think this is clear... it at least makes it quite clear that it's threaded. I don't think we need to change it

data_diff/diff_tables.py
erezsh: done

erezsh:
I have some ideas on how to make the thread-count predictable, but I'm not even sure if it's necessary, since in 99% of the cases it won't be more than a few hundreds. (and max database connections are already predictable)

sirupsen (approved):
LGTM, nice and simple!

I think having lots of green threads in Python is acceptable. I think it's ok to not worry more about it for now. It'd make things more difficult to reason about I think.

data_diff/__main__.py
```diff
@@ -37,6 +37,7 @@
 @click.option("-d", "--debug", is_flag=True, help="Print debug info")
 @click.option("-v", "--verbose", is_flag=True, help="Print extra info")
 @click.option("-i", "--interactive", is_flag=True, help="Confirm queries, implies --debug")
+@click.option("--threads", default=None, help="Number of threads to use. 1 means no threading. Auto if not specified.")
```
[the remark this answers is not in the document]
erezsh: Good point! I've seen it around. Any idea what it stands for?

erezsh:
They are regular threads, not green threads. But they'll spend most of their time idling.

PR #32, opened 2022-05-19
performance: remove all full index scans, divide keyspace instead

sirupsen:
@erezsh putting this up for review a little earlier than I am comfortable with, because I can use some feedback now as I polish this up for merge! I'm quite aware it's not there yet (see my inline comments) 

This is a fairly large change. I explained to you the basic idea already of splitting `min(id)...max(id)` into equal-sized segments, which I think can work well. Overall, we need to ban _any_ full-index or table scans in the code-base, which you'll see I've done in a few places.

In a further iteration, we can allow setting the `--bisect-column`.

With `--bisection-threshold 10000 --bisection-factor 32` on `rating <-> rating_del1` with 25m rows in Postgres:

* **Before:** 74s
* **After:** 16s (4-5x faster)


data_diff/diff_tables.py
```diff
@@ -97,37 +120,84 @@ def segment_by_checkpoints(self, checkpoints: List[DbKey]) -> List["TableSegment
...
         """Using new() creates a copy of the instance using 'replace()', and makes sure the cache is reset"""
         return self.replace(_count=None, _checksum=None, **kwargs)
 
+    def __repr__(self):
+        return f"{type(self.database).__name__}/{', '.join(self.table_path)}"
+
+    def query_start_key_and_end_key(self) -> Tuple[int, int]:
+        """Query database for minimum and maximum key. This is used for setting
+        the boundaries of the initial, full table table segment."""
+        select = self._make_select(columns=[f"min({self.key_column})", f"max({self.key_column})"])
```
sirupsen, line 133: I will create a proper `Min` / `Max` in `sql.py` 👍🏻

data_diff/diff_tables.py
```diff
@@ -97,37 +120,84 @@ def segment_by_checkpoints(self, checkpoints: List[DbKey]) -> List["TableSegment
...
+        start_key = res[0] or 1
+        # TableSegments are always exclusive the last key:
+        #   (1..250) => # WHERE i >= 1 AND i < 250
+        # Thus, for the very last segment (which is the one where these
+        # aren't automatically set!) -- we have to add 1.
+        end_key = res[1] + 1 if res[1] else 1
+
+        return (start_key, end_key)
+
+    def compute_checksum_and_count(self):
```
sirupsen, line 145: The idea here is that we shouldn't ad-hoc calculate these things, it should be quite intentional. I don't love this complexity, but I think it's fairly clear with the errors. Unfortunately performance does increase complexity sometimes...

data_diff/diff_tables.py
```diff
@@ -160,7 +229,7 @@ class TableDiffer:
     """
 
     bisection_factor: int = 32  # Into how many segments to bisect per iteration
-    bisection_threshold: int = 1024**2  # When should we stop bisecting and compare locally (in row count)
+    bisection_threshold: int = 10000  # When should we stop bisecting and compare locally (in row count)
```
sirupsen, line 232: I think 1M is _way_ too high when we are running this against databases far away from the machine. 10k seems more sensible. We can run some benchmarks in the future, but I think this is a safer default

tests/test_diff_tables.py
```diff
@@ -61,8 +61,12 @@ def test_basic(self):
         differ = TableDiffer(10, 100)
         a = TableSegment(self.connection, ("a",), "id", "datetime")
         b = TableSegment(self.connection, ("b",), "id", "datetime")
-        assert a.count == 6
-        assert b.count == 5
+        differ.set_initial_start_key_and_end_key(a, b)
+        a.compute_checksum_and_count()
+        b.compute_checksum_and_count()
```
sirupsen, line 66: In tests this stuff is really ugly... I can create a helper function for it

data_diff/diff_tables.py
```diff
@@ -176,49 +245,74 @@ def diff_tables(self, table1: TableSegment, table2: TableSegment) -> DiffResult:
...
-            assert count1 == sum(s.count for s in segmented1)
-            assert count2 == sum(s.count for s in segmented2)
+        # [1..249, 250..499, 500..749, 750..1000]
+        segmented1 = table1.segment_by_checkpoints(checkpoints)
+        segmented2 = table2.segment_by_checkpoints(checkpoints)
 
         # Compare each pair of corresponding segments between table1 and table2
         for i, (t1, t2) in enumerate(safezip(segmented1, segmented2)):
-            logger.info(". " * level + f"Diffing segment {i+1}/{len(segmented1)} of size {t1.count} and {t2.count}")
+            logger.info(". " * level + f"Diffing segment {i+1}/{len(segmented1)} keys={t1.start_key}..{t1.end_key-1}")
```
sirupsen, line 313: We have to remove all the counts because they're too expensive, that's why we're just writing the keyspace here. I want to do that before computing the checksum in case it's slow, but we could move it after

data_diff/diff_tables.py
```diff
@@ -97,37 +120,84 @@ def segment_by_checkpoints(self, checkpoints: List[DbKey]) -> List["TableSegment
...
     @property
     def checksum(self) -> int:
         if self._checksum is None:
-            self._checksum = (
-                self.database.query(self._make_select(columns=[Checksum(self._relevant_columns)]), int) or 0
-            )
+            # Get the count in the same index pass. Much cheaper than doing it
+            # separately.
+            select = self._make_select(columns=[Count(), Checksum(self._relevant_columns)])
+            result = self.database.query(select, Tuple)
```
sirupsen, line 197: Probably needs some error handling

data_diff/diff_tables.py
```diff
@@ -68,27 +71,47 @@ def get_values(self) -> list:
...
-        assert ratio > 1
-        skip = f"mod(idx, {ratio}) = 0"
-        select = self._make_select(table=Enum(self.table_path, order_by=self.key_column), where=skip)
-        return self.database.query(select, List[int])
+    def choose_checkpoints(self, bisection_factor: int) -> List[DbKey]:
+        """
+        Choose `bisection_factor - 1` (because of start and end) checkpoints in
+        the keyspace.
 
-    def find_checkpoints(self, checkpoints: List[DbKey]) -> List[DbKey]:
```
sirupsen, line 79: We completely get rid of `find_checkpoints` since we're just creating these segments. It doesn't matter if the ids actually exist

data_diff/diff_tables.py
sirupsen: The relationship here between `@property` and this pre-computation function may be too weird. Is it Python-ish to do it this way, or would it be more natural to use well-named methods?

data_diff/sql.py
```diff
@@ -80,6 +80,12 @@ def compile(self, parent_c: Compiler):
         if self.where:
             select += " WHERE " + " AND ".join(map(c.compile, self.where))
 
+        if self.where_or:
```
erezsh, line 83: What's this?

erezsh (changes requested):
Partial review.

sirupsen:
Feedback from Erez on Slack:

* Create benchmarks with large gaps to ensure performance here is good compared to `master` (I think it will be!)
* More thorough tests, e.g. when `bisection_threshold < gap` and other edge-cases

data_diff/sql.py
sirupsen: Ah yeah this is a relic left over from the first iteration where I found checkpoints still with this method, will remove

erezsh:
I managed to reproduce the behavior that was troubling me. Here are the benchmarks, when running on 25m records (postgres):

| Table A | Table B | branch | time |
|-|-|-|-|
| gap3 | gap3_update | master | 137.70 seconds
| gap3 | gap3_update | no-scans | 425.50 seconds

Just x3 times slower isn't THAT bad, but this effect would become much worse with a bigint id, or a smaller threshold.

Here's the script to generate the "gap3" tables:

```go
const table rating_gap3 = rating

// create_indices(rating_gap1)
// create_indices(rating_gap2)
create_indices(rating_gap3)
commit()

table rating_gap3 {
    userid: int
    movieid: int
    rating: float
    timestamp: int
}

rating_gap3[id == 1000] update {id: 2147483548}
run_sql("INSERT INTO rating_gap3(id, userid, movieid, rating, timestamp) VALUES (2047483548, 1, 1, 5.0, 27)")
commit()

const table rating_gap3_update0001p = rating_gap3
rating_gap3_update0001p[random() < 0.000001] update {timestamp: timestamp + 1}
commit()
```

PR #34, opened 2022-05-25
Added colors to diff output

erezsh:


erezsh:
Btw you can ignore whitespace in github too: https://github.com/datafold/data-diff/pull/34/files?diff=split&w=1

data_diff/__main__.py
```diff
@@ -1,129 +1,138 @@
...
+    "Example: --min-age=5min ignores rows from the last 5 minutes. "
+    f"\nValid units: {UNITS_STR}",
+)
+@click.option("--max-age", default=None, help="Considers only rows younger than specified. See --min-age.")
+@click.option("-s", "--stats", is_flag=True, help="Print stats instead of a detailed diff")
+@click.option("-d", "--debug", is_flag=True, help="Print debug info")
+@click.option("-v", "--verbose", is_flag=True, help="Print extra info")
+@click.option("-i", "--interactive", is_flag=True, help="Confirm queries, implies --debug")
+@click.option(
+    "-j", "--threads", default=None, help="Number of threads to use. 1 means no threading. Auto if not specified."
```
sirupsen, line 47: Isn't the default 1 thread? What does auto mean? I think it should be 1 so that the output is nice and expected in the default case. It's also far safer on a relational database where much more could really have an impact (especially with the current queries).

Sorry I know this is not strictly related here, but I must've missed this on the other PR. I thought I read for this when I reviewed the threading PR

sirupsen (approved):
LGTM

data_diff/__main__.py
erezsh: Let's discuss it on slack.

PR #36, opened 2022-05-25
Improve the -j option; set default value to 1

erezsh:


data_diff/__main__.py
```diff
@@ -44,7 +44,12 @@
 @click.option("-v", "--verbose", is_flag=True, help="Print extra info")
 @click.option("-i", "--interactive", is_flag=True, help="Confirm queries, implies --debug")
 @click.option(
-    "-j", "--threads", default=None, help="Number of threads to use. 1 means no threading. Auto if not specified."
+    "-j",
+    "--threads",
+    default="1",
+    help="Number of worker threads to use per database. Default=1. "
```
sirupsen, line 50: Would it be worth adding some more information here about tuning of this?

```
"Number of worker threads to use per database to execute queries simultaneously. Default = 1"
"A higher number will increase performance, but take more capacity from your database."
"Tune this while monitoring your database."
```

data_diff/__main__.py
```diff
@@ -44,7 +44,12 @@
...
+    "'auto' will use the maximum amount recommended by Python. "
```
sirupsen, line 51: What does "recommended by Python" mean? 👂🏻

sirupsen (commented):
Just some questions and suggestions, but if this block you tomorrow, you can merge and address in follow-ups

data_diff/__main__.py
erezsh: It usually means equal to the number of cores on your local computer.

erezsh: (I know it's not necessarily the best option for database workers, but at least it keeps it within a reasonable range)

data_diff/__main__.py
erezsh: It's definitely good information to provide. The question is should this be part of the help text, or somewhere in the docs?

sirupsen: I think at least the first line in the help text, if you think it's all too long.

data_diff/__main__.py
sirupsen: Yeah it's a bit weird... I think we should just remove `auto` then 🤷🏻 I think it's more confusing than valuable.

erezsh: Fair enough

sirupsen (approved):

PR #37, opened 2022-05-26
Performance: Implementation of Simon's algorithm, minimizing the amount of scan-queries

erezsh:


erezsh:
Initial implementation.

Tasks left:

- `set(relevant_columns)`. Btw, why did you use sort? @sirupsen 
- Add more tests
- Add prepare_db_gap
- better log prints

data_diff/diff_tables.py
```diff
@@ -162,7 +161,7 @@ class TableDiffer:
     """
 
     bisection_factor: int = 32  # Into how many segments to bisect per iteration
-    bisection_threshold: int = 1024**2  # When should we stop bisecting and compare locally (in row count)
+    bisection_threshold: int = 1024 * 16  # When should we stop bisecting and compare locally (in row count)
```
sirupsen, line 164: Sure, don't have a huge opinion on this until we benchmark this on some remote hosts, but I feel much better about it than 1M

erezsh:
Np. There's just a small semantic, that we assume the first element is always the key column. Maybe it was just luck that so far `id` ended up first after a sort.

data_diff/diff_tables.py
```diff
@@ -124,13 +105,31 @@ def _relevant_columns(self) -> List[str]:
             + list(self.extra_columns)
         )
 
+    def count(self) -> Tuple[int, int]:
+        return self.database.query(self._make_select(columns=[Count()]), int)
```
sirupsen, line 109: The reason I didn't include this was because I was worried that someone might accidentally call this in `diff_tables`, which slows the whole thing down... This might be a work injury from working on big codebases where this _will_ happen, so I'm OK if you call me over-cautious and keep it. You can also make the argument that if we have a proper benchmark suite that's the better fix.

It just feels a bit weird to me that this method is _only_ there for testing

data_diff/diff_tables.py
```diff
@@ -183,52 +182,47 @@ def diff_tables(self, table1: TableSegment, table2: TableSegment) -> DiffResult:
         if self.bisection_factor < 2:
             raise ValueError("Must have at least two segments per iteration (i.e. bisection_factor >= 2)")
 
-        return self._diff_tables(table1, table2)
+        logger.info(
+            f"Diffing tables | segments: {self.bisection_factor}, bisection threshold: {self.bisection_threshold}."
```
sirupsen, line 186: I think we should put the key range in here too as I had 👀

data_diff/diff_tables.py
```diff
@@ -183,52 +182,47 @@ def diff_tables(self, table1: TableSegment, table2: TableSegment) -> DiffResult:
...
+        )
 
-    def _diff_tables(self, table1, table2, level=0, segment_index=None, segment_count=None):
-        count1, count2 = self._precalc_attr_threaded("count", [table1, table2])
+        key_ranges = self._threaded_call("query_key_range", [table1, table2])
+        mins, maxs = zip(*key_ranges)
+
+        # We add 1 because our ranges are exclusive of the end (like in Python)
+        start_key = min(mins)
+        end_key = max(maxs) + 1
```
sirupsen, line 194: I think the previous version where this is spelled is a bit more readable and explicit, but I don't feel too strongly.

data_diff/diff_tables.py
```diff
@@ -183,52 +182,47 @@ def diff_tables(self, table1: TableSegment, table2: TableSegment) -> DiffResult:
...
-        checksum1, checksum2 = self._precalc_attr_threaded(
-            "checksum", [table1, table2]
-        )  # Calculate checksum in parallel
+        if max_rows is None:
+            # We can be sure that row_count <= max_rows
+            max_rows = table1.end_key - table1.start_key
 
-        if checksum1 == checksum2:
-            return  # No differences
+        if max_rows == 0:
```
sirupsen, line 208: Why is this condition necessary?

data_diff/diff_tables.py
```diff
@@ -183,52 +182,47 @@ def diff_tables(self, table1: TableSegment, table2: TableSegment) -> DiffResult:
...
-            logger.info(
-                f"Diffing tables of size {table1.count} and {table2.count} | segments: {self.bisection_factor}, bisection threshold: {self.bisection_threshold}."
-            )
+    def _bisect_and_diff_tables(self, table1, table2, level=0, max_rows=None):
+        assert table1.is_bounded and table2.is_bounded
 
-        checksum1, checksum2 = self._precalc_attr_threaded(
-            "checksum", [table1, table2]
-        )  # Calculate checksum in parallel
+        if max_rows is None:
```
sirupsen, line 204: Ah, I didn't see that you passed `max(count1, count2)` here when recursing. Took me a second. I thought for a bit that you'd done away with using `count` here beyond at level > 0.

I think the recursion would be clearer if you just passed `table1.end_key - table1.start_key` from the caller for the first iteration 👍🏻

sirupsen (commented):
Definitely quite a bit simpler! Thanks for picking this up!

data_diff/diff_tables.py
erezsh: I think of it as API for the users too. Once they create a TableSegment, they can count to see how much it catches.

I'm not too worried that we'll call it by accident. Also it's a method now and not a property, which signifies (int terms of semantics) that it might be an expensive operation.

But I don't have a strong feeling about this, so I'm open to change my mind.

data_diff/diff_tables.py
erezsh: It feels easier to read for me. I guess it's subjective.

Also, this version is a preparation for more handling more than 2 tables.

data_diff/diff_tables.py
erezsh: Optimization. So we don't `get_values()` on an empty segment.

data_diff/diff_tables.py
sirupsen: Let's just keep it

data_diff/__main__.py
```diff
@@ -30,7 +30,7 @@
 @click.option("-c", "--columns", default=[], multiple=True, help="Names of extra columns to compare")
 @click.option("-l", "--limit", default=None, help="Maximum number of differences to find")
 @click.option("--bisection-factor", default=32, help="Segments per iteration")
-@click.option("--bisection-threshold", default=1024**2, help="Minimal bisection threshold")
+@click.option("--bisection-threshold", default=1024 * 16, help="Minimal bisection threshold")
```
sirupsen, line 33: Could we reference a constant here instead so we don't duplicate the default across here and the Python API? 👍🏻

data_diff/diff_tables.py
```diff
@@ -239,12 +236,31 @@ def _diff_tables(self, table1, table2, level=0, segment_index=None, segment_coun
         for res in self._thread_map(list, diff_iters):
             yield from res
 
+    def _diff_tables(self, table1, table2, level=0, segment_index=None, segment_count=None):
+        logger.info(
+            ". " * level
+            + f"Diffing segment {segment_index}/{segment_count}, key-range: {table1.start_key:x}..{table2.end_key:x}"
```
sirupsen, line 242: I think the size is really useful too, even if it's not actually the # of rows in some cases.

data_diff/diff_tables.py
```diff
@@ -239,12 +236,31 @@ def _diff_tables(self, table1, table2, level=0, segment_index=None, segment_coun
...
+        )
+
+        (count1, checksum1), (count2, checksum2) = self._threaded_call("count_and_checksum", [table1, table2])
+
+        if not (count1 or count2):
```
sirupsen, line 247: I always forget that 0 is falsey in Python... I would probably prefer `count1 == 0 or count2 == 0`

data_diff/diff_tables.py
```diff
@@ -239,12 +236,31 @@ def _diff_tables(self, table1, table2, level=0, segment_index=None, segment_coun
...
+            logger.warn(
+                "Uneven distribution of keys detected. "
```
sirupsen, line 249: We're not really detecting an uneven distribution of keys, but rather very large key segments where we're not seeing any rows, due to gaps.

data_diff/diff_tables.py
```diff
@@ -239,12 +236,31 @@ def _diff_tables(self, table1, table2, level=0, segment_index=None, segment_coun
...
+                "For better performance, we recommend to increase the bisection-threshold."
```
sirupsen, line 250: 👍🏻 Good. I think it might also be worth highlighting in the warning that it will still work.

By the way, will this log to stderr automatically? If you're piping this into a file where you want all the rows that are different to go, then we need to make sure these warnings aren't printed there :)

dev/prepare_db_gaps.pql
```diff
@@ -0,0 +1,78 @@
+// This is a Preql file, used for setting up a database for developement and testing
```
sirupsen, line 1: Can we include this one in the other file? Or you think this is such an edge-case it's not really worth populating in the common dev environment?

dev/prepare_db_gaps.pql
```diff
@@ -0,0 +1,78 @@
...
+
+run_sql("UPDATE rating_gap1 SET id = id * 1000 + 25000000 where 100000 < id and id <= 500000 ")
+run_sql("UPDATE rating_gap2 SET id = cast(id*0.1*id as int) + 26000000 WHERE 10 < id and id < 100000")
+run_sql("INSERT INTO rating_gap3(id, userid, movieid, rating, timestamp) VALUES (2047483548, 1, 1, 5.0, 27)")
+commit()
+
+print "Create more tables"
+const table rating_gap1_update0001p = rating_gap1
+const table rating_gap2_update0001p = rating_gap2
+const table rating_gap3_update0001p = rating_gap3
```
sirupsen, line 73: Do these tables have the indices? Are those cloned too in this case?

dev/prepare_db_gaps.pql
```diff
@@ -0,0 +1,78 @@
...
+    movieid: int
+    rating: float
+    timestamp: int
+}
+
+rating_gap3[id == 1000] update {id: 2147483548}
+
+run_sql("UPDATE rating_gap1 SET id = id * 1000 + 25000000 where 100000 < id and id <= 500000 ")
+run_sql("UPDATE rating_gap2 SET id = cast(id*0.1*id as int) + 26000000 WHERE 10 < id and id < 100000")
+run_sql("INSERT INTO rating_gap3(id, userid, movieid, rating, timestamp) VALUES (2047483548, 1, 1, 5.0, 27)")
```
sirupsen, line 67: Can you add the comments back on what "type" of gaps these all are?

data_diff/diff_tables.py
```diff
@@ -99,38 +96,46 @@ def segment_by_checkpoints(self, checkpoints: List[DbKey]) -> List["TableSegment
...
+    def count(self) -> Tuple[int, int]:
+        return self.database.query(self._make_select(columns=[Count()]), int)
+
+    def count_and_checksum(self) -> Tuple[int, int]:
+        start = time.time()
+        count, checksum = self.database.query(
+            self._make_select(columns=[Count(), Checksum(self._relevant_columns)]), tuple
         )
+        duration = time.time() - start
+        if duration > RECOMMENDED_CHECKSUM_DURATION:
```
sirupsen, line 120: Hm... so at first I know I suggested 1s, but now I'm thinking that on e.g. Snowflake it's actually kind of OK and even expected for something to run 5-10s, but on MySQL/Postgres 5-10s is really bad... Yeah, I don't feel too strongly, just feel good that we have _something_ in place.

data_diff/diff_tables.py
```diff
@@ -99,38 +96,46 @@ def segment_by_checkpoints(self, checkpoints: List[DbKey]) -> List["TableSegment
...
+            logger.warn(f"Checksum is taking longer than expected ({duration:.2f}s). ")
```
sirupsen, line 121: We need to offer a suggestion here: Increase the `bisection_factor`

We could have this same warning for `get_values` if it takes more than this duration, then `bisection_threshold` might be too high.

sirupsen (commented):
I can't wait to get this in! We're so close 😍

data_diff/__main__.py
erezsh: Could we? Let me check if such a thing is possible ;)

data_diff/diff_tables.py
erezsh: The gaps exist due to an uneven distribution..

data_diff/diff_tables.py
erezsh: It's better to have a suggestion, but I don't know what it is yet. I need to run some tests to know what would actually help.

data_diff/diff_tables.py
sirupsen: Yeah ok sorry lol 

What I mean to say I think it's a bit more user-friendly to talk about it in terms of gaps having few keys...

data_diff/diff_tables.py
sirupsen: Increasing the `bisection_factor` means that the key ranges are going to be smaller, which means that individual queries should be faster 👍🏻

dev/prepare_db_gaps.pql
erezsh: Yep, I think it's only worth testing if you mess with the basic algorithm, and most devs won't have anything to do with it.

data_diff/diff_tables.py
erezsh: I don't know why we should be worried about that, as it won't increase the total run-time.

Actually, now that I think of it, I'm not sure what's the point of this warning.

sirupsen: The reason we need to be worried about it is that it's bad form to do queries on a relational databases that take several seconds. They take up a lot of capacity.

data_diff/diff_tables.py
sirupsen: Why you you writing these in hex form? Pretty hard to read

CleanShot 2022-05-27 at 12 33 45@2x

sirupsen:
Didn't mean to approve, was meant as another comment

erezsh:
In what way is it hard to read?

sirupsen:
@erezsh I want to see the actual id integers. I don't understand why you'd want them in hex

sirupsen:
Just remembered: For `--stats` we need to still not do the `count()`. We instead need to tally up the total count from the `level=0` segments and save it for later.

erezsh:
Here's a thought: We can make `TableSegment.count()` a use bisection on itself and query the table in threads. We already have the mechanism for it, and it could be a nice API for the users, while still useful for tests.

data_diff/diff_tables.py
```diff
@@ -239,12 +245,36 @@ def _diff_tables(self, table1, table2, level=0, segment_index=None, segment_coun
...
+        if count1 == 0 and count2 == 0:
+            logger.warn(
+                "Uneven distribution of keys detected. (big gaps in the key column). "
+                "For better performance, we recommend to increase the bisection-threshold."
+            )
+            assert checksum1 is None and checksum2 is None
+            return
+
+        if level == 1:
+            self.stats["table1_count"] = self.stats.get("table1_count", 0) + count1
```
sirupsen, line 267: Wouldn't it make more sense to set this on the `TableSegment`? `table1` is kind of an internal vocabulary, this would be a bit hard to support as an external API. It's OK with me if you don't want to worry about this right now, and want to address that with the Python API, and just get it merged.

I think this goes to your idea of a parallelized count, too. To me a "parallelized count" would be calling `diff_tables`, but just not doing `Checksum()` in the query—only the count. That could be a cool idea 🤷🏻

sirupsen (approved):
Looking great, I think this is about ready to merge!!

data_diff/diff_tables.py
erezsh: We want `TableSegment` to be a reusable structure. For example, to be used by two diffs at the same time. If we accumulate the sum over there, we'll have two threads in a race condition that is almost certain to happen.

So I think it's right to accumulate inside of `TableDiffer`. We can, afterwards, store the result in TableSegment. This is something that's easy to add later too.

A parallelized count would use the same base mechanism as diff_tables, but it would run a different sequence, since it's only 1 table, so a lot of the diff operations like checksums and comparisons aren't necessary. It makes sense for it to be in a different function.

Anyway, I think both of these, if we want them, belong in a new PR.

PR #39, opened 2022-05-27
Added support for 'a.b.table_name' syntax

erezsh:


data_diff/__main__.py
```diff
@@ -99,8 +101,8 @@ def main(
         logging.error("Error while parsing age expression: %s" % e)
         return
 
-    table1 = TableSegment(db1, (table1_name,), key_column, update_column, columns, **options)
-    table2 = TableSegment(db2, (table2_name,), key_column, update_column, columns, **options)
+    table1 = TableSegment(db1, parse_table_name(table1_name), key_column, update_column, columns, **options)
+    table2 = TableSegment(db2, parse_table_name(table2_name), key_column, update_column, columns, **options)
```
sirupsen, line 105: I think to your previous point, to make the Python API easier to use, what do you think about doing this inside of TableSegment?

erezsh: I think most users shouldn't create a TableSegment. They can instead call something like "create_table" which will do all the init and parsing necessary.

sirupsen: OK, as long as we use that here too

PR #40, opened 2022-05-27
readme: update for new algorithm, add db table, add primer

sirupsen:
@erezsh I've re-worked the README:

* Brief summary of how it works in intro
* Add table with performance and support and sample connection strings (for ease of use)
* Re-work the technical explanation for the new algorithm and diagrams

Would love your feedback.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ please file an issue and we'll help you out ASAP!**
 rows across two different databases.
 
 * 🪢 Verifies across [many different databases][dbs] (e.g. Postgres -> Snowflake)
-* 🔍 Outputs diff of rows in detail
+* 🔍 Outputs [diff of rows](#example-output) in detail
 * 🚨 Simple CLI/API to create monitoring and alerts
 * 🔥 Verify 25M+ rows in less than 10s
 * ♾️  Works for tables with 10s of billions of rows
@@ -22,6 +22,17 @@ This approach has similar performance to `count(*)` when there are few/no
 changes, but is able to output each differing row (and it might even [be
 faster][perf]). By pushing the compute into the databases, it's _much_ faster
 than querying for and comparing every row.
+
+## Table of Contents
+
+- [Common use-cases](#common-use-cases)
+- [Example output](#example-output)
+- [Supported Databases](#supported-databases)
+- [How to install](#how-to-install)
+- [How to use](#how-to-use)
+- [Technical Explanation](#technical-explanation)
+- [Performance Considerations](#performance-considerations)
+- [Development Setup](#development-setup)
 
 ## Common use-cases
 
@@ -136,11 +147,10 @@ Options:
   - `-d` or `--debug` - Print debug info
   - `-v` or `--verbose` - Print extra info
   - `-i` or `--interactive` - Confirm queries, implies `--debug`
-  - `--max-age` - Considers only rows younger than specified. See `--min-age`.
-                  Useful for specifying replication lag.
   - `--min-age` - Considers only rows older than specified.
-                  Example: `--min-age=5min` considers only rows from the last 5 minutes.
+                  Example: `--min-age=5min` ignores rows from the last 5 minutes.
                   Valid units: `d, days, h, hours, min, minutes, mon, months, s, seconds, w, weeks, y, years`
+  - `--max-age` - Considers only rows younger than specified. See `--min-age`.
   - `--bisection-factor` - Segments per iteration. When set to 2, it performs binary search.
   - `--bisection-threshold` - Minimal bisection threshold. i.e. maximum size of pages to diff locally.
   - `-j` or `--threads` - Number of worker threads to use per database. Default=1.
@@ -300,18 +310,19 @@ If you pass `--stats` you'll see e.g. what % of rows were different.
   Explained in the [technical explanation][tech-explain].
 * If there are very large gaps in your table, e.g. 10s of millions of
   continuous rows missing, then **data-diff** may perform poorly doing lots of
-  queries that for rows that do not exist (see [technical
+  queries for ranges of rows that do not exist (see [technical
   explanation][tech-explain]). There are various things we could do to optimize
   the algorithm for this case with complexity that has not yet been introduced,
   please open an issue.
-* The fewer columns you verify, the faster **data-diff** will be. On one extreme
-  you can verify every column, on the other you can verify _only_ `updated_at`,
-  if you trust it enough, or `id` if you're just interested in presence. You can
-  do also do a hybrid where you verify `updated_at` and the most critical value,
-  e.g a money value in `amount` but not verify a large serialized column like
-  `json_settings`.
+* The fewer columns you verify (passed with `--columns`), the faster
+  **data-diff** will be. On one extreme you can verify every column, on the
+  other you can verify _only_ `updated_at`, if you trust it enough. You can also
+  _only_ verify `id` if you're interested in only presence, e.g. to detect
+  missing hard deletes. You can do also do a hybrid where you verify
+  `updated_at` and the most critical value, e.g a money value in `amount` but
+  not verify a large serialized column like `json_settings`.
 
-## Development Setup
+# Development Setup
 
 The development setup centers around using `docker-compose` to boot up various
 databases, and then inserting data into them.
--- a/data_diff/__main__.py
+++ b/data_diff/__main__.py
@@ -41,12 +41,10 @@ def parse_table_name(t):
     "--min-age",
     default=None,
     help="Considers only rows older than specified. "
-    "Example: --min-age=5min considers only rows from the last 5 minutes. "
+    "Example: --min-age=5min ignores rows from the last 5 minutes. "
     f"\nValid units: {UNITS_STR}",
 )
-@click.option("--max-age", default=None,
-              help="Considers only rows younger than specified. See --min-age."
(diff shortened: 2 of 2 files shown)


erezsh:
I think you get too quickly into specifics. Instead, overviews should start abstract, and when possible, get into specifics layer by layer.

Personally, it would be much easier for me to read a technical analysis of the algorithm and the main features, than a story full of examples where I have to read several pages to get the whole picture.

Also, I'm not sure about the table you put. Did we test snowflake more than Redshift? I'm not aware of it.

sirupsen:
> I think you get too quickly into specifics. Instead, overviews should start abstract, and when possible, get into specifics layer by layer.

The introduction at the top of the README is the abstract overview. The technical explanation is exactly how I've explained it to users, and it seems to resonate very well. In my experience, example-walkthroughs like this are very powerful, and most users glance over/don't read thinks that are highly abstract.

We need an outsider to judge though, as that's your insider opinion against mine :) I've asked in the channel.

> Also, I'm not sure about the table you put. Did we test snowflake more than Redshift? I'm not aware of it.

Yep. One company we ran it with did Postgres <-> Snowflake for a sizable table.

README.md
```diff
@@ -1,52 +1,85 @@
...
-
-We plan to add more, including NoSQL, and even APIs like Shopify! We're very
-open to contributions on this front.
+**data-diff** is a command-line tool and Python library to efficiently validate
+whether two different databases are in sync, e.g. Postgres and Redshift. The
+common use-case is to verify replication between two datastores with e.g.
+Fivetran, Debezium, .. and fix missing rows and/or alert on missing/incorrect
+data so you know whether something is wrong before your customer.
+
+**data-diff** doesn't pull down and compare every row, because it is extremely
```
AlexanderOtavka, line 12: It kinda sounds like you are calling data-diff slow, lol.

```suggestion
**data-diff** doesn't pull down and compare every row, because that would be extremely
```

erezsh:
Don't get me wrong, I think the walkthrough is great. I should have led with that. But I think it should come after a more brief and technical explanation, which I think is valuable to have for the more technical audience.

README.md
```diff
@@ -80,50 +113,136 @@ Options:
...
+                    ┃                  ┃
+                    ┃ checksum queries ┃
+                    ┃                  ┃
+                  ┌─┻──────────────────┻────┐
+                  │        data-diff        │
+                  └─────────────────────────┘
+```
 
-    2. If their size is below the threshold, diff them locally and print the results.
+Now **data-diff** will start running `--threads=1` queries in parallel that
```
AlexanderOtavka, line 173: Don't forget to add `--threads` to the options reference up above in the "how to use" section.

AlexanderOtavka (commented):
Very nice explanation, I love the ASCII art.

AlexanderOtavka:
I think your approach leading with the abstract followed by an example and setup instructions is good. When I open up a README my first two questions are 1. what does this do? and 2. how do I use it?

README.md
sirupsen: Yep, we're just revising the wording for this in another PR — where we'll do this 👍🏻 

We want `--threads=1` to mean "1 simultaneous query per database", today it doesn't really mean that...

sirupsen:
> Don't get me wrong, I think the walkthrough is great. I should have led with that. But I think it should come after a more brief and technical explanation, which I think is valuable to have for the more technical audience.

But wouldn't this duplicate the intro, which I think kinda does that? 👂🏻 I can write a paragraph to briefly explain it again if you prefer!

erezsh:
@sirupsen I was also talking about the overview. I think the original overview, while for sure it needs improvement, was much clearer and more informative.

sirupsen:
@erezsh updated according to our session. Let me know what you think

README.md
```diff
@@ -1,52 +1,112 @@
...
+* 🪢 Verifies across [many different databases][1] (e.g. Postgres -> Snowflake)
+* 🔍 Outputs diff of rows in detail
+* 🚨 Simple CLI/API to create monitoring and alerts
+* 🔥 Verify 25M+ rows in less than 10s
+* ♾️  Works for tables with 10s of billions of rows
+
+**data-diff** splits the table into smaller segments, then checksums each
+segment in both databases. When the checksums for a segment aren't equal, it
+will further divide that segment into yet smaller segments, cheksumming those
+until it gets to the offending row(s). See [Technical Explanation][2] for more
```
erezsh, line 18: Re "offending", I would stick to something less picturesque, like differing, or changed.

README.md
```diff
@@ -1,52 +1,112 @@
...
+details.
+
+This approach has similar performance to `count(*)`, but is able to output each
```
erezsh, line 21: Only true when there are no changes. Also, maybe worth mentioning threading here?

README.md
```diff
@@ -1,52 +1,112 @@
...
+differing row. By pushing the compute into the databases, it's _much_ faster
+than querying for and comparing each row.
```
erezsh, line 23: maybe "every row" instead of "each row"

README.md
```diff
@@ -65,66 +125,157 @@ Usage: `data-diff DB1_URI TABLE1_NAME DB2_URI TABLE2_NAME [OPTIONS]`
...
   - `-l` or `--limit` - Maximum number of differences to find (limits maximum bandwidth and runtime)
   - `-s` or `--stats` - Print stats instead of a detailed diff
   - `-d` or `--debug` - Print debug info
   - `-v` or `--verbose` - Print extra info
   - `-i` or `--interactive` - Confirm queries, implies `--debug`
+  - `--max-age` - Considers only rows younger than specified. See `--min-age`.
+                  Useful for specifying replication lag.
   - `--min-age` - Considers only rows older than specified.
-                  Example: `--min-age=5min` ignores rows from the last 5 minutes.
+                  Example: `--min-age=5min` considers only rows from the last 5 minutes.
```
erezsh, line 139: It's ignore, not considers.

It looks like it's really confusing, so maybe we should re-think the name or phrasing?

data_diff/__main__.py
```diff
@@ -41,10 +41,12 @@ def parse_table_name(t):
     "--min-age",
     default=None,
     help="Considers only rows older than specified. "
-    "Example: --min-age=5min ignores rows from the last 5 minutes. "
+    "Example: --min-age=5min coniders only rows from the last 5 minutes. "
```
erezsh, line 44: Same thing, this change isn't right.

erezsh (commented):
Looks a lot better imho!

I didn't re-read the tutorial yet, but here's the comments I have so far.

README.md
sirupsen: Hmmmm yeah we definitely need to rethink it, but isn't this right?

Drew a diagram below... Here the `--min-age` is the `id >= now() -5min` and `--max-age` is `id <= now() - 2min`?

```
┌ ─ ─ ┬─────┬─────┬─────┬─────┬ ─ ─ ┬ ─ ─ ┬ ─ ─ ┐
 11:59│12:00│12:01│12:02│12:03│12:04 12:05 12:06
└ ─ ─ ┴──▲──┴─────┴─────┴──▲──┴ ─ ─ ┴ ─ ▲ ┴ ─ ─ ┘
         │                 │            │
         │                 │            │
         │                 │            │
  ┌──────┴───────┐    ┌────┴─────────┐  │ ┌─────────────┐
  │ --min-age=5m │    │ --max-age=2m │  └─┤Present time │
  └──────────────┘    └──────────────┘    └─────────────┘
```

sirupsen:
@erezsh fixed it up based on your review, let me know if you think this is ready to go:)

README.md
erezsh: No, min-age at this time means anything before 12.

And a min-age that's bigger than the max-age shouldn't be allowed.

When you go back in time the "time value" or timestamp get smaller, that's true, but the age actually gets bigger. (like someone born in the 80s is "bigger" than someone from the 90s, despite the number being smaller)

The "age" of 12:00 is 5 mins. The age of 12:01 is 4 mins, which is below 5, aka the minimum age, so it's out.

The "age" of 11:00 is an hour and 5 mins, which is a bigger age than the minimum age.

Does it make sense?

sirupsen: Yeah, but I mean... I drew it, and I'm still confused 😓  

But I think we need to change it to `--min-updated-key` and `--max-updated-key` which would line up with the drawing above

erezsh: But then it's not very clear that the value is relative to now. It sounds more like an absolute value..

erezsh:
Okay

README.md
sirupsen: Yeah, but I think it's clearer then if you do `--min-updated -5min` or something.

erezsh:
@sirupsen Icon isn't showing

image

sirupsen:
I guess the Emoji isn't in Windows yet
I don't have a great other suggestion... 

🖇️ 
🧳
📑

README.md
AlexanderOtavka: Hi Y'all, end user here.  I think the semantics of “age” matter a lot.  The source of confusion is that “age” is implicitly negative, but we think about it as a positive number/absolute value, and so inverts the conventional meaning of “min” and “max” when thinking about timestamps.  Your readme documents `--min-age` being a lower bound on “age” and an upper bound on “timestamp” (matching my intuition), but the implementation does the reverse.

Regardless, docs should match readme, and ideally should match user intuition as well.

PR #41, opened 2022-05-27
presto: add docker-compose setup

sirupsen:
This adds a `docker-compose` setup for PrestoSQL v 340.

You can run this with `docker-compose up presto`.

It's accessible via:

```python
import prestodb
prestodb.dbapi.connect(host="127.0.0.1", user="presto").cursor().execute('SELECT * FROM system.runtime.nodes')
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ RUN apt-get update && apt-get install -y \
 RUN pip install --no-cache-dir poetry==1.1.13
 COPY . /app
 WORKDIR /app
-# For now while we are in heavy development we install the latest with Poetry 
+# For now while we are in heavy development we install the latest with Poetry
 # and execute directly with Poetry. Later, we'll move to the released Pip package.
-RUN poetry install -E preql -E mysql -E pgsql -E snowflake
+RUN poetry install
 ENTRYPOINT ["poetry", "run", "python3", "-m", "data_diff"]
--- a/dev/Dockerfile.prestosql.340
+++ b/dev/Dockerfile.prestosql.340
@@ -1,0 +1,25 @@
+FROM openjdk:11-jdk-slim-buster
+
+ENV PRESTO_VERSION=340
+ENV PRESTO_SERVER_URL=https://repo1.maven.org/maven2/io/prestosql/presto-server/${PRESTO_VERSION}/presto-server-${PRESTO_VERSION}.tar.gz
+ENV PRESTO_CLI_URL=https://repo1.maven.org/maven2/io/prestosql/presto-cli/${PRESTO_VERSION}/presto-cli-${PRESTO_VERSION}-executable.jar
+ENV PRESTO_HOME=/opt/presto
+ENV PATH=${PRESTO_HOME}/bin:${PATH}
+
+WORKDIR $PRESTO_HOME
+
+RUN set -xe \
+    && apt-get update \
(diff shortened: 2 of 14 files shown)


Dockerfile
```diff
@@ -5,7 +5,7 @@ RUN apt-get update && apt-get install -y \
 RUN pip install --no-cache-dir poetry==1.1.13
 COPY . /app
 WORKDIR /app
-# For now while we are in heavy development we install the latest with Poetry 
+# For now while we are in heavy development we install the latest with Poetry
 # and execute directly with Poetry. Later, we'll move to the released Pip package.
-RUN poetry install -E preql -E mysql -E pgsql -E snowflake
+RUN poetry install
```
sirupsen, line 10: This installs all dev dependencies anyway; not necessary to do the extras anymore.

dev/Dockerfile.prestosql.340
```diff
@@ -0,0 +1,25 @@
+FROM openjdk:11-jdk-slim-buster
```
sirupsen, line 1: Borrowed from [this buddy's Github](https://github.com/vimagick/dockerfiles/tree/master/prestosql)

erezsh:
@sirupsen The docker works, but I still can't change the db. Any advice?

```python
cur.execute('create table tpch.tiny.a(x int)')
cur.fetchall()
  ...
prestodb.exceptions.PrestoUserError: PrestoUserError(type=USER_ERROR, name=NOT_SUPPORTED, message="This connector does not support creating tables", query_id=20220530_152518_00004_mverw)
```

sirupsen:
@erezsh I added `dev/presto-conf/standalone/catalog/postgresql.properties` with the appropriate configuration for connecting to the `docker-compose` managed PostgreSQL. 👍🏻 

If you want to use your own, local PostgreSQL you'd need to update the username/password. In addition, you'll need to map the port from `localhost` into the container, doing what's described in this [Stackoverflow answer](https://stackoverflow.com/questions/31324981/how-to-access-host-port-from-docker-container). But I suggest you just switch to the one in `docker-compose` and import the 1M table one in there with the instructions from the README.

For me locally this script:

```python
import prestodb
cur = prestodb.dbapi.connect(host="127.0.0.1", user="presto").cursor()
print(list(cur.execute('SHOW SCHEMAS FROM postgresql')))
print(list(cur.execute('SELECT * FROM postgresql.public.rating LIMIT 10')))
```

Yields:

```
[['information_schema'], ['pg_catalog'], ['public']]
[[1, 1, 296, 5.0, 1147880044], [2, 1, 306, 3.5, 1147868817], [3, 1, 307, 5.0, 1147868828], [4, 1, 665, 5.0, 1147878820], [5, 1, 899, 3.5, 1147868510], [6, 1, 1088, 4.0, 1147868495], [7, 1, 1175, 3.5, 1147868826], [8, 1, 1217, 3.5, 1147878326], [9, 1, 1237, 5.0, 1147868839], [10, 1, 1250, 4.0, 1147868414]]
```

Because Presto is just a query layer on top, you should be able to just use the existing `rating*` tables 👍🏻

erezsh:
Okay, it seems to be working! Thanks for the help

sirupsen:
@erezsh cool, good with me merging then?

erezsh:
I think so.

PR #42, opened 2022-05-30
New API draft

erezsh:


erezsh:
Common usage will look like this:

```python
from data_diff import diff_tables, TableRef   # will change to datadiff in another PR

table1 = TableRef('postgres:///', 'Rating')
table2 = TableRef('postgres:///', 'Rating_del1')
diff_tables(table1, table2, update_column="timestamp", bisection_factor=64)
```

data_diff/__init__.py
```diff
@@ -1,9 +1,86 @@
...
+    DbKey,
+    DbTime,
+    parse_table_name,
+)
 
 
-def create_source(db_uri: str, table_name: str, key_column: str, extra_columns: Tuple[str, ...] = ()):
-    db = connect_to_uri(db_uri)
-    return TableSegment(db, (table_name,), key_column, tuple(extra_columns))
+class TableRef:
```
sirupsen, line 15: Remind me the reasoning why we introduce a `TableRef` on top of `TableSegment`? 👂🏻

data_diff/__init__.py
```diff
@@ -1,9 +1,86 @@
-from typing import Tuple
+from typing import Tuple, Iterator
```
sirupsen, line 1: I know you're not doing this yet because you want it to be finalized, and not conflict with the other README PR, but I think we need to add some examples to the README, as well as a link to the Pydocs 👍🏻

data_diff/__init__.py
```diff
@@ -1,9 +1,86 @@
...
+    # Name of updated column, which signals that rows changed (usually updated_at or last_update)
+    update_column: str = None,
+    # Extra columns to compare
+    extra_columns: Tuple[str, ...] = (),
+    # Start/end key_column values, used to restrict the segment
+    start_key: DbKey = None,
+    end_key: DbKey = None,
+    # Start/end update_column values, used to restrict the segment
+    min_time: DbTime = None,
+    max_time: DbTime = None,
```
sirupsen, line 44: Hmmm, should we mirror `update_column` and call it `min_update` and `max_update`?

data_diff/__init__.py
```diff
@@ -1,9 +1,86 @@
...
+    def create_table_segment(self, **kwargs):
+        return TableSegment(self.db, parse_table_name(self.table_name), **kwargs)
+
+
+def diff_tables(
+    table1: TableRef,
+    table2: TableRef,
+    *,
+    # Name of the key column, which uniquely identifies each row (usually id)
+    key_column: str = "id",
```
sirupsen, line 34: I've been thinking about whether we should call this the "bisection_column", but I think that might be leaking too many implementation details... Just important to note that this could be `tenant_id`, `created_at`, ... whatever. As long as it's numeric and has a decent distribution.

data_diff/__init__.py
```diff
@@ -1,9 +1,86 @@
...
+    max_time: DbTime = None,
+    # Into how many segments to bisect per iteration
+    bisection_factor: int = DEFAULT_BISECTION_FACTOR,
+    # When should we stop bisecting and compare locally (in row count)
+    bisection_threshold: int = DEFAULT_BISECTION_THRESHOLD,
+    # Enable/disable threaded diffing. Needed to take advantage of database threads.
+    threaded: bool = True,
+    # Maximum size of each threadpool. None = auto. Only relevant when threaded is True.
+    # There may be many pools, so number of actual threads can be a lot higher.
+    max_threadpool_size: int = None,
```
sirupsen, line 53: Why isn't this exactly symmetric with the CLI options?

data_diff/__init__.py
```diff
@@ -1,9 +1,86 @@
...
+    ]
+
+    differ = TableDiffer(
+        bisection_factor=bisection_factor,
+        bisection_threshold=bisection_threshold,
+        debug=debug,
+        threaded=threaded,
+        max_threadpool_size=max_threadpool_size,
+    )
+    return differ.diff_tables(*segments)
```
sirupsen, line 86: Can the iterator yield the results as they come in, instead of blocking? Or is that what this does?

data_diff/__init__.py
sirupsen: The other thing is tests (which I know you didn't add for the same reason; want to finalize it first)

data_diff/diff_tables.py
```diff
@@ -247,8 +272,7 @@ def _bisect_and_diff_tables(self, table1, table2, level=0, max_rows=None):
 
     def _diff_tables(self, table1, table2, level=0, segment_index=None, segment_count=None):
```
sirupsen, line 273: While you're in here, could you output `key-range` and `size` for the full table above too? I think it's OK to wait for that `min/max` on both ends first.

data_diff/__init__.py
```diff
@@ -1,9 +1,86 @@
...
+        self.db = connect_to_uri(db_uri)
+        self.table_name = table_name
+
+    def create_table_segment(self, **kwargs):
+        return TableSegment(self.db, parse_table_name(self.table_name), **kwargs)
+
+
+def diff_tables(
+    table1: TableRef,
+    table2: TableRef,
```
sirupsen, line 31: Do you want to make this an iterator/list right away?

data_diff/__init__.py
erezsh: TableSegment is complicated to construct, and it is used more as an internal structure. 99% of the users don't care about start_key, and if they give a start_time they'll only want to give it once, instead of duplicate it for each TableSegment.

Users can still use it, but we should offer a simple and easy API for the common use case.

data_diff/__init__.py
erezsh: Does it make it clearer? I think it's okay is it.

data_diff/__init__.py
erezsh: created_at isn't guaranteed to be unique, so it won't work well as a key.

data_diff/__init__.py
erezsh: This way gives the programmer more control. The CLI is a simplification, to make it easier.

Or if you mean why the default isn't 1, I just missed it and will fix.

data_diff/__init__.py
erezsh: Yes, that's what the iterator does. The pipeline isn't as smooth as it could be, due to threads, but that's something we can improve in the future.

data_diff/__init__.py
erezsh: No, I just think it's probably going to be some day.

Either way, using the args like `(db, table, db, table)` isn't good practice imho, and leads to confusion.

data_diff/__init__.py
sirupsen: I think `update` makes it clearer, yeah.

data_diff/__init__.py
sirupsen: Hmmm does the new algorithm care if they're unique though?

sirupsen:
It's looking pretty good, let me try to play a bit around with the API in an almost-empty file and I'll see if any other feedback comes out of that!

data_diff/__init__.py
erezsh: I think the bisection will still work, as long as it's mostly unique. But it will mess with the stats feature.

data_diff/__main__.py
```diff
@@ -31,7 +29,7 @@ def parse_table_name(t):
 @click.option("-t", "--update-column", default=None, help="Name of updated_at/last_updated column")
 @click.option("-c", "--columns", default=[], multiple=True, help="Names of extra columns to compare")
 @click.option("-l", "--limit", default=None, help="Maximum number of differences to find")
-@click.option("--bisection-factor", default=32, help="Segments per iteration")
+@click.option("--bisection-factor", default=DEFAULT_BISECTION_FACTOR, help="Segments per iteration")
```
sirupsen, line 32: Remember to import it

sirupsen:
Looking really good! 🚀 

Some thoughts from playing around with a script locally (thanks for the sample!):

1. In the documentation, I think we should have an example line that sets the logging level to `INFO` so you can see the verbose output the first time you run it. I think that's very satisfying!
2. I think we need to support counts and optionally stats out of the gate.... Maybe the simplest way to do this is that the helper helps create `TableDiffer`, you call `diff_tables` yourself, and then we can allow them to peek into the `stats` property? I think there's a case here for the helper to help you create the object, rather than run the whole thing 👀 
3. I'm still not sure I understand the value of the `TableRef` class. You mentioned it's because doing this kind of thing in the initializer in `TableSegment` is bad form... I'm not sure I agree. Just seems like a common question why we have both which will cause more confusion than I think automatically parsing these tables will 🤷🏻 
4. In your example of `rating` as table1 and `rating_del1` as table2, why does the API return `[('+', (12500048,))]`? I would've expected the diff to be "left-to-right", that is, for it to return `[('+', (12500048,))]`.
5. When running this against `rating` and `rating_001p` it's somehow not working? Just gives me an empty list? Seems like it's not recursing? I ran the CLI, and it worked fine 🤔

data_diff/__init__.py
erezsh: Changed to min_update / max_update

erezsh:
@sirupsen Addressing your points:

1. Sure, why not.
2. What stats specifically do you think we should provide? (other than getting the table count)
3. I don't want anyone to create a TableSegment on their own, because there's a lot they can get wrong. There won't be any confusion, because no one should access the TableSegment class directly. Will it help if I rename `TableRef` to `connect_to_table()`, or something like that?
4. The two samples you wrote are exactly the same. (`[('+', (12500048,))]`)
5. I don't know. Can you include the code that doesn't work for you?

sirupsen:
> What stats specifically do you think we should provide? (other than getting the table count)

Yeah, we can probably get away with just # of rows, and # of different rows. I guess the latter could even just be from `len()`.

> Will it help if I rename TableRef to connect_to_table(), or something like that?

I think a function that returns a properly crated `TableSegment` is more appropriate here, yeah. 

> The two samples you wrote are exactly the same. ([('+', (12500048,))])

I meant that I expected it to be `([('-', (12500048,))])`. That represents a "diff" from left-to-right in my opinion, and is more consistent with how e.g. `git diff` works, no?

> I don't know. Can you include the code that doesn't work for you?

Works great now! For my own reference, this is the code, but works just fine now!

```python
from data_diff import diff_tables, TableRef   # will change to datadiff in another PR
import logging

logging.basicConfig(level=logging.INFO)
table1 = TableRef('postgres://postgres:Password1@localhost/postgres', 'rating')
table2 = TableRef('postgres://postgres:Password1@localhost/postgres', 'rating_update001p')
print(list(diff_tables(table1, table2, update_column="timestamp", debug=True, bisection_factor=64)))
```

data_diff/__init__.py
```diff
@@ -1,9 +1,86 @@
...
+    tables = [table1, table2]
+    segments = [
+        t.create_table_segment(
+            key_column=key_column,
+            update_column=update_column,
+            extra_columns=extra_columns,
+            min_key=min_key,
+            max_key=max_key,
+            min_updated=min_updated,
+            max_updated=max_updated,
```
sirupsen, line 74: `min_update` — this needs a test 😉

data_diff/__init__.py
```diff
@@ -1,9 +1,86 @@
...
+def diff_tables(
+    table1: TableRef,
+    table2: TableRef,
+    *,
+    # Name of the key column, which uniquely identifies each row (usually id)
+    key_column: str = "id",
+    # Name of updated column, which signals that rows changed (usually updated_at or last_update)
+    update_column: str = None,
+    # Extra columns to compare
+    extra_columns: Tuple[str, ...] = (),
```
sirupsen, line 38: I kind of think this should just be "columns" in the user-facing API, but I don't feel too strongly

data_diff/__main__.py
```diff
@@ -31,7 +35,7 @@ def parse_table_name(t):
 @click.option("-t", "--update-column", default=None, help="Name of updated_at/last_updated column")
```
sirupsen, line 35: What do you think about exposing `--min-update` and `--max-update` here too while we're at it?

data_diff/diff_tables.py
```diff
@@ -31,33 +32,50 @@ def split_space(start, end, count):
...
-        if self.start_key is not None:
-            yield Compare("<=", str(self.start_key), self.key_column)
-        if self.end_key is not None:
-            yield Compare("<", self.key_column, str(self.end_key))
+        if self.min_key is not None:
+            yield Compare("<=", str(self.min_key), self.key_column)
+        if self.max_key is not None:
+            yield Compare("<", self.key_column, str(self.max_key))
 
     def _make_update_range(self):
```
sirupsen, line 74: The last confusion caused by our relative parsing could likely be fixed by erroring with a helpful error message if `min_update > max_update`

sirupsen (commented):
Code looks good to me! Only smaller suggestions at your discretion.

README update with an example, a link to Pydocs, and then I think this is good to go!

data_diff/__init__.py
erezsh: I think `columns` might imply that it overrides `key_column` or `update_column`, or that you need to specfy them twice.

data_diff/__main__.py
erezsh: Maybe? Let's leave that for another PR though. Always easy to add.

erezsh:
@sirupsen Ok I am *blocked* from merging because there is an unresolved conversation and github is glitching.

I don't understand why I have to deal with these issues when I'm the one writing 95% of the commits. Whoever is playing with the settings isn't helping.

PR #43, opened 2022-05-30
Bugfixes and better types

erezsh:


data_diff/database.py
```diff
@@ -77,7 +77,7 @@ def query(self, sql_ast: SqlOrStr, res_type: type):
             logger.info(self._query(explained_sql))
             answer = input("Continue? [y/n] ")
             if not answer.lower() in ["y", "yes"]:
-                os.exit(1)
+                sys.exit(1)
```
sirupsen, line 80: lol thx, I guess it worked and I just didn't read the error properly 🤦🏻

data_diff/database.py
```diff
@@ -361,7 +361,7 @@ def connect_to_uri(db_uri: str, thread_count: Optional[int] = 1) -> Database:
     elif scheme == "bigquery":
         return BigQuery(dsn.host, path)
     elif scheme == "redshift":
-        return Redshift(dsn.host, dsn.port, path, dsn.user, dsn.password)
+        return Redshift(dsn.host, dsn.port, path, dsn.user, dsn.password, thread_count=thread_count)
```
sirupsen, line 364: What about bigquery, snowflake, ..?

erezsh: They currently don't use a thread-pool. 

Because they don't need connection pooling, and connecting to them is slow, it will probably be better to use a different base class that makes threads but uses the same connection. I'd just want to test them on bigquery/snowflake, and so far we focused on postgres/mysql.

So we need to be sure that

1) Using parallel requests is really faster for cloud providers as well

2) Using the same connection really works without issues

So it was just on the back burner. If you think it's important, I can prioritize this (though it's just performance).

erezsh: All tests point to yes - our current approach is good enough.

PR #44, opened 2022-05-31
Presto

erezsh:


sirupsen (commented):
LGTM, what's the performance like? I presume that Presto basically does do a `SELECT *` and then applies those transformations in memory :/

sirupsen (approved):

erezsh:
Performance is bad for memory, which might have affect the runtime too. But either way, it wasn't too bad. Maybe 3 times slower than direct postgres.

sirupsen:
👍🏻 good to merge for me

PR #54, opened 2022-06-09
Normalize Fields (Replacing PR #49)

erezsh:
Replacing PR #49 

- Query database for table schema and parse it
- Always convert to normalized-repr before checksum or download
  - Support for dates with precision and timezone (needs more testing)
- Negotiate minimal mutual precision when diffing two columns (needs more testing)


data_diff/database.py
```diff
@@ -163,8 +250,20 @@ def close(self):
 _CHECKSUM_BITSIZE = CHECKSUM_HEXDIGITS << 2
 CHECKSUM_MASK = (2**_CHECKSUM_BITSIZE) - 1
 
+DEFAULT_PRECISION = 6
+
+TIMESTAMP_BASE_LEN = 20
```
sirupsen, line 255: Can you add a comment with how it's this bit:

```ruby
"2022-06-03 12:24:35."
```

data_diff/database.py
```diff
@@ -173,7 +272,9 @@ def __init__(self, host, port, database, user, password, *, thread_count):
     def create_connection(self):
         postgres = import_postgres()
         try:
-            return postgres.connect(**self.args)
+            c = postgres.connect(**self.args)
+            # c.cursor().execute("SET TIME ZONE 'UTC'")
```
sirupsen, line 276: I assume you just used this to debug? I think it's fair game for us to change settings on the connection, as long as they don't affect the whole database. It's so much easier to normalize everything to UTC per session if that's easier, than try to wrangle equality otherwise 🤯

data_diff/database.py
```diff
@@ -186,6 +287,12 @@ def md5_to_int(self, s: str) -> str:
     def to_string(self, s: str):
         return f"{s}::varchar"
 
+    def normalize_value_by_type(self, value, coltype: ColType) -> str:
```
sirupsen, line 290: The relationship between the 'compiler' in `sql.py` and what's in the Database drivers themselves is starting to blur a bit for me 👀 

Do you think it'd be worth moving some of this driver-specific logic increasingly in there, so you can call e.g. `Normalize()` with a Database as another argument?

data_diff/database.py
```diff
@@ -186,6 +287,12 @@ def md5_to_int(self, s: str) -> str:
...
+        if isinstance(coltype, TemporalType):
+            timestamp = self.to_string(f"{value}::timestamp({coltype.precision})")
+            return f"RPAD({timestamp}, {TIMESTAMP_BASE_LEN + coltype.precision}, '0')"
```
sirupsen, line 293: So MySQL below does right-padding by default to the precision? 👍🏻 

FWIW, I think in the abstract PyDoc (since it's where people will look when implementing a new driver) for `#normalize_value_by_type` I think it's worth explaining in detail how we convert everything to a string before checksumming (to avoid relying on e.g. different implementations of floats, varints, ...). 

Here, I'd also start the list of types that we expect some canonical representation for, in this case, with Datetime/Timestamp we always expect: `2022-06-03 12:24:35.000000` (if precision is 6).

data_diff/database.py
```diff
@@ -261,11 +380,45 @@ def quote(self, s: str):
     def to_string(self, s: str):
         return f"cast({s} as varchar(1024))"
 
+    def select_table_schema(self, path: DbPath) -> str:
+        if len(path) > 1:
+            raise ValueError("Unexpected table path for oracle")
+        (table,) = path
+
+        return (
+            f"SELECT column_name, data_type, 9 as datetime_precision, data_precision as numeric_precision"
```
sirupsen, line 389: CleanShot 2022-06-10 at 10 26 19@2x

Default for Oracle seems to be 6, not 9

data_diff/database.py
```diff
@@ -261,11 +380,45 @@ def quote(self, s: str):
...
+            f" FROM USER_TAB_COLUMNS WHERE table_name = '{table.upper()}'"
+        )
+
+    def normalize_value_by_type(self, value, coltype: ColType) -> str:
+        if isinstance(coltype, PrecisionType):
+            return f"to_char(cast({value} as timestamp), 'YYYY-MM-DD HH24:MI:SS.FF{coltype.precision or ''}')"
```
sirupsen, line 395: Why do you need to convert to `to_char` here? Doesn't the wrapping `concat` do that automatically?

data_diff/diff_tables.py
```diff
@@ -142,7 +159,7 @@ def count_and_checksum(self) -> Tuple[int, int]:
         """Count and checksum the rows in the segment, in one pass."""
         start = time.time()
         count, checksum = self.database.query(
-            self._make_select(columns=[Count(), Checksum(self._relevant_columns)]), tuple
+            self._make_select(columns=[Count(), Checksum(self._relevant_columns_repr)]), tuple
```
sirupsen, line 162: What do you think about in `debug` mode, we output `concat` value generated by the database? Seems really useful to debug the types of issues you've run into. So you see something like for `id=2`, `time=2020-01-01 00:00:00.000000`:

```
CHECKSUMMED VALUE: 22020-01-01 00:00:00.000000
```

data_diff/database.py
```diff
@@ -371,6 +558,15 @@ def md5_to_int(self, s: str) -> str:
     def to_string(self, s: str):
         return f"cast({s} as string)"
 
+    def select_table_schema(self, path: DbPath) -> str:
+        schema, table = self._normalize_path(path)
+        return super().select_table_schema((schema.upper(), table.upper()))
+
+    def normalize_value_by_type(self, value, coltype: ColType) -> str:
+        if isinstance(coltype, TemporalType):
+            return f"{value}::timestamp({coltype.precision})::text"
```
sirupsen, line 567: This is by far the most elegant!

data_diff/database.py
```diff
@@ -321,14 +481,36 @@ def _query(self, sql_code: str):
...
+        schema, table = self._normalize_path(path)
+
+        return (
+            f"SELECT column_name, data_type, 6 as datetime_precision, 6 as numeric_precision FROM {self.dataset}.INFORMATION_SCHEMA.COLUMNS "
+            f"WHERE table_name = '{table}' AND table_schema = '{schema}'"
+        )
+
+    def normalize_value_by_type(self, value, coltype: ColType) -> str:
+        if isinstance(coltype, PrecisionType):
+            return self.to_string(f'FORMAT_TIMESTAMP("%F %H:%M:%E6S", {value})')
```
sirupsen, line 503: Don't you need to interpolate the precision value here, if the source is < 6? If it's not possible, at least raising an exception like for Redshift 👍🏻

data_diff/database.py
```diff
@@ -261,11 +380,45 @@ def quote(self, s: str):
...
+        return UnknownColType(type_repr)
+
 
 class Redshift(Postgres):
     def md5_to_int(self, s: str) -> str:
         return f"strtol(substring(md5({s}), {1+MD5_HEXDIGITS-CHECKSUM_HEXDIGITS}), 16)::decimal(38)"
 
+    def normalize_value_by_type(self, value, coltype: ColType) -> str:
+        if coltype.precision != 6:
+            raise TypeError(f"Bad precision for Redshift: {coltype})")
```
sirupsen, line 419: 👍🏻  Consider adding a comment here with the bug that's preventing this

data_diff/diff_tables.py
```diff
@@ -227,6 +244,24 @@ def diff_tables(self, table1: TableSegment, table2: TableSegment) -> DiffResult:
         if self.bisection_factor < 2:
             raise ValueError("Must have at least two segments per iteration (i.e. bisection_factor >= 2)")
 
+        table1, table2 = self._threaded_call("with_schema", [table1, table2])
+        for c in table1._relevant_columns:
```
sirupsen, line 248: Yes!!!!! ❤️ 🚀 

This is such an incredible seed for a lot of beautiful warnings and exceptions as we find 'incompatibilities' between the schemas. We'll (you especially) will become _experts_ in all of the incompatibilities between various types, between various databases. For some people who are doing a migration, the wealth of knowledge encoded in the warnings might be _more_ valuable to some than the actual checking! This will be gold. 🏅 

Might be worth throwing into its own method as you polish this up for merge.

data_diff/diff_tables.py
```diff
@@ -227,6 +244,24 @@ def diff_tables(self, table1: TableSegment, table2: TableSegment) -> DiffResult:
...
+            # Update schemas to minimal mutual precision
+            col1 = table1._schema[c]
+            col2 = table2._schema[c]
+            if isinstance(col1, PrecisionType):
+                if not isinstance(col2, PrecisionType):
+                    raise TypeError(f"Incompatible types for column {c}:  {col1} <-> {col2}")
+
+                min_precision = min(col1.precision, col2.precision)
+                table1._schema[c] = col1.replace(precision=min_precision)
+                table2._schema[c] = col2.replace(precision=min_precision)
```
sirupsen, line 263: 😍 😍 😍 😍 😍 😍 

If `min_precision` isn't equal to both precisions, I think you should output an information warning here!

data_diff/database.py
```diff
@@ -66,12 +70,77 @@ def _query_conn(conn, sql_code: str) -> list:
...
+class TimestampTZ(TemporalType):
+    pass
+
+
+class Datetime(TemporalType):
+    pass
+
+
+@dataclass
+class UnknownColType(ColType):
```
sirupsen, line 99: What do you think about before launch we create basic types for _all_ types we talked about having strong support for: ints, datetime, floats + currency, bools, and strings. Then, if you are using a column of _any_ other type we print a bunch of warnings in `_parse_type` or in `diff_tables` where we say (if the source and target are different databases):

"Column `settings` is type `json`. While it _may_ work, we haven't tested this type."

Then we could add a type support table like this to the README:

| Category          | Postgres | MySQL | Oracle | Snowflake | Redshift |
|-------------------|----------|-------|--------|-----------|----------|
| int               | ✅        | ✅     | ✅      | ✅         | ✅        |
| datetime          | ✅        | ✅     | ✅      | ✅         | ✅        |
| text              | ✅        | ✅     | ✅      | ✅         | ✅        |
| float             | ⌛        | ⌛     | ⌛      | ⌛         | ⌛        |
| decimal, currency | ⌛        | ⌛     | ⌛      | ⌛         | ⌛        |
| bool              | ⌛        | ⌛     | ⌛      | ⌛         | ⌛        |
| date, time        | ⌛        | ⌛     | ⌛      | ⌛         | ⌛        |
| json              | ⌛        | ⌛     | ⌛      | ⌛         | ⌛        |
| array             | ⌛        | ⌛     | ⌛      | ⌛         | ⌛        |

My tests will then cover each cell in that matrix against each other cell in that row.

data_diff/diff_tables.py
```diff
@@ -71,6 +73,13 @@ def __post_init__(self):
         if self.min_update is not None and self.max_update is not None and self.min_update >= self.max_update:
             raise ValueError("Error: min_update expected to be smaller than max_update!")
 
+    def with_schema(self) -> "TableSegment":
+        "Queries the table schema from the database, and returns a new instance of TableSegmentWithSchema."
```
sirupsen, line 77: It's not actually a different class though, no?

Why does this instantiate a new segment when e.g. the count and checksum don't?

tests/test_diff_tables.py
```diff
@@ -151,7 +154,7 @@ def test_diff_small_tables(self):
         )
         self.preql.commit()
         diff = list(self.differ.diff_tables(self.table, self.table2))
-        expected = [("-", (2, datetime.datetime(2022, 1, 1, 0, 0)))]
+        expected = [("-", ("2", time + ".000000"))]
```
sirupsen, line 157: The ints are strings now because we're converting everything to text for every database, yeah?

tests/test_normalize_fields.py
```diff
@@ -0,0 +1,109 @@
...
+
+from data_diff.database import BigQuery, MySQL, Snowflake, connect_to_uri, Oracle, DEFAULT_PRECISION
+from data_diff.sql import Select
+from data_diff import database as db
+
+from .common import CONN_STRINGS
+
+
+DATE_TYPES = {
+    db.Postgres: ["timestamp({p}) with time zone", "timestamp({p}) without time zone"],
```
sirupsen, line 15: How exactly do the time zone and without timezone compare correctly? I guess it works because you always drop the timezone bit?

Should we include the timezone in the string comparison if _both_ types have a timezone, to ensure they're the same? I don't think we have to do that now, and I know it's a bit tricky for e.g. MySQL where we have to figure out what timezone MySQL is in to get it (but it's doable).

I feel it's more important that we have support for floats and have proper types for all "type categories" we support more so than this though.

data_diff/database.py
```diff
@@ -66,12 +70,77 @@ def _query_conn(conn, sql_code: str) -> list:
...
+class ColType:
+    pass
+
+
+@dataclass
+class PrecisionType(ColType):
+    precision: Optional[int]
+
+
+class TemporalType(PrecisionType):
```
sirupsen, line 82: I think `TemporalType` also would then include `Time`, `Date`, and `Interval` eventually, which are a bit different. We'll proper need another layer of class between this over-arching time type and `Timestamp`, `TimestampTZ`, and `Datetime`

data_diff/database.py
```diff
@@ -66,12 +70,77 @@ def _query_conn(conn, sql_code: str) -> list:
...
+
+class Timestamp(TemporalType):
+    pass
+
+
+class TimestampTZ(TemporalType):
+    pass
+
+
+class Datetime(TemporalType):
```
sirupsen, line 94: Should we just _only_ do `Timestamp` and `TimestampTZ` for all of our representation to simplify? Do we really need `Datetime` too?

It seems most of the new databases who aren't too concerned with SQL standards just call everything a timestamp, and that the old distinction of `Timestamp` being about a current point-in-time and `Datetime` being for e.g. birthdays and stuff doesn't hold anymore.

data_diff/database.py
erezsh: Yeah np

data_diff/database.py
erezsh: > starting to blur a bit for me 

Yeah, I agree. It's just a lot of db-specific code, and very few generic SQL code that works everywhere. It makes it feel like the design is wrong. But it's the databases that are wrong ;)

> Do you think it'd be worth moving some of this driver-specific logic increasingly in there,

I'll do what I can, but I think it's inevitable that Database will provide quite a bit of db-specific SQL code

> Normalize() with a Database as another argument

I don't think it's a good idea. You'd never use `[ Normalize(x, db1), Normalize(x, db2) ]`, so we shouldn't make it possible to express.

data_diff/database.py
erezsh: Link please

data_diff/database.py
erezsh: `to_char()` does the date formatting.

Of course it does. It's called to_char(), what else could it do?

data_diff/diff_tables.py
erezsh: Maybe, yeah.

data_diff/database.py
erezsh: Hmm yeah probably. I had bigquery disabled because of the issues with snowflake, I forgot to test it 😞

data_diff/database.py
erezsh: Technically we can also support precision = 3. But I don't know if it's worth the effort.

data_diff/diff_tables.py
erezsh: Good point!

data_diff/database.py
erezsh: I think that's a lot of work, so I don't think we'll make it before launch.

We can try though.

tests/test_diff_tables.py
erezsh: Yeah.

tests/test_normalize_fields.py
erezsh: > Should we include the timezone in the string comparison if both types have a timezone, to ensure they're the same?

I don't know. But I think it's not a good idea.

data_diff/database.py
erezsh: Yeah, maybe. Not sure how to call it though.

data_diff/database.py
erezsh: Datetime is just the database type, that we get from querying the schema. We already convert it to timestamp no matter what.

tests/common.py
```diff
@@ -1,17 +1,38 @@
...
+TEST_MYSQL_CONN_STRING: str = "mysql://mysql:Password1@localhost/mysql"
+TEST_POSTGRES_CONN_STRING: str = None
+TEST_SNOWFLAKE_CONN_STRING: str = None
+TEST_BIGQUERY_CONN_STRING: str = None
+TEST_REDSHIFT_CONN_STRING: str = None
+TEST_ORACLE_CONN_STRING: str = None
+
 
 try:
     from .local_settings import *
```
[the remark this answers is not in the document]
erezsh: Nope, tests are now failing for it because it's not set up for github actions. Reverting. 👍🏻

data_diff/database.py
sirupsen: https://docs.oracle.com/cd/B19306_01/server.102/b14225/ch4datetime.htm#i1006081

data_diff/database.py
sirupsen: LOL

data_diff/database.py
sirupsen: Ok, we can discuss later if we need to change the design a bit. I think this'll happen more and more with the other types.. 😅

data_diff/database.py
sirupsen: Ya, a nice-to-have, post-launch is fine too.

data_diff/diff_tables.py
sirupsen: Fair

data_diff/database.py
sirupsen: Yeah, but then why have another class for it?

erezsh: Because we might want to make this distinction in the future, or add datetime/timestamp-specific behaviors. And saying "datetime" is Timestamp is just going to cause weird bugs for someone in the future.

PR #55, opened 2022-06-13
Support databricks

pik94:
Add test support of Databricks connector.

### DIFF
--- a/data_diff/databases/databricks.py
+++ b/data_diff/databases/databricks.py
@@ -61,7 +61,7 @@ class Databricks(Database):
         return f"`{s}`"
 
     def md5_to_int(self, s: str) -> str:
-        return f"conv(substr(md5({s}), {1+MD5_HEXDIGITS-CHECKSUM_HEXDIGITS}), 16, 10)"
+        return f"cast(conv(substr(md5({s}), {1+MD5_HEXDIGITS-CHECKSUM_HEXDIGITS}), 16, 10) as decimal(38, 0))"
 
     def to_string(self, s: str) -> str:
         return f"cast({s} as string)"


pyproject.toml
```diff
@@ -38,9 +38,11 @@ snowflake-connector-python = { version = "*", optional = true }
 [tool.poetry.dev-dependencies]
 mysql-connector-python = "*"
 preql = "^0.2.14"
+pandas = "^1.3"
```
erezsh, line 41: Why is this here?

erezsh:
Also, needs to be rebased over `master`; there are conflicts to resolve. (sorry, we're at a fast moving dev stage)

sirupsen:
@erezsh I think we'll also need the rounding support in here, no?

pyproject.toml
```diff
@@ -23,7 +23,7 @@ classifiers = [
 packages = [{ include = "data_diff" }]
 
 [tool.poetry.dependencies]
-python = "^3.7"
+python = "^3.7.1"
```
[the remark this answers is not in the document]
pik94: Without it poetry install old versions of databricks-sql-connector (<2.0.2) and pandas (which is used by databricks), and it fails when running the diff. Increase of a patch version helped, and higher version of databricks and pandas are installed.

pyproject.toml
pik94: With a lower version, the databricks connector fails because an old version of pandas (<1.3.*) when using. But maybe it can be removed because if databricks has a good version (2.0.2), pandas should have an appropriate good version as well. I'll check if it can be removed.

pik94:
> Also, needs to be rebased over `master`; there are conflicts to resolve. (sorry, we're at a fast moving dev stage)

@erezsh , I resolved but I see that a new version failed because it could not compare object of [ColumnName](https://github.com/datafold/data-diff/blob/master/data_diff/sql.py#L52) class. I've added a fix in this PR for it.

erezsh:
Hi, I'm sorry but a lot has changed in the last week.

`Database` classes now also need to "normalize" the values, based on the column type. For example, dates need to have the form of `YYYY-MM-DD HH:MM:SS.FFFFFF`. They also need to provide table schema, so we know the column types.

We plan to make the 0.1 release sometime next week. I believe the pace of internal changes will slow down after that.

erezsh:
We've released v0.1 and things should be a bit more stable from now on.

data_diff/databases/connect.py
```diff
@@ -132,6 +137,12 @@ def connect_to_uri(db_uri: str, thread_count: Optional[int] = 1) -> Database:
             kw["password"] = dsn.password
     kw = {k: v for k, v in kw.items() if v is not None}
 
+    if scheme == "databricks":
+        # dsn user - access token
+        # sdn password - http path (starting with /)
+
+        return cls(http_path=dsn.password, access_token=dsn.user, server_hostname=dsn.hostname, **kw)
```
erezsh, line 144: I don't know how I feel about putting the `http_path` as the password.

Can you tell me why you chose this, instead of adding it to the path, or as a query argument?

pik94: > adding it to the path

http_path starts with `/` symbol so adding it to the path leads to an error when parsing. We can just notify users no to use the slash in their path but I bet what the first run will be always failed because from Databricks UI http_path is copied with the slash so people might be confused during their first attempt of diffing.

> a query argument

I assumed that query params in terms of usage look like additional i.e. not necessarily so people might skip it for the first run, and it'll lead to an error as well.

So, my concern it just not to disappoint people in their first attempt of diffing.

data_diff/databases/databricks.py
```diff
@@ -0,0 +1,125 @@
...
+                if issubclass(type_cls, Integer):
+                    row = (row.COLUMN_NAME, row.TYPE_NAME, None, None, 0)
+
+                elif issubclass(type_cls, Float):
+                    numeric_precision = math.ceil(row.DECIMAL_DIGITS / math.log(2, 10))
+                    row = (row.COLUMN_NAME, row.TYPE_NAME, None, numeric_precision, None)
+
+                elif issubclass(type_cls, Decimal):
+                    # TYPE_NAME has a format DECIMAL(x,y)
+                    items = row.TYPE_NAME[8:].strip(')').split(',')
```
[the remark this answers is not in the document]
pik94: Np, changed it.)

data_diff/databases/connect.py
erezsh: We can provide a good error, telling users that this parameter is necessary.

I think it's less confusing than using the password. Also, we're going to remove the password whenever we print the configuration to the logs. That means the http_path will appear as `****`.

P.S. we can also ensure that it's formatted with only one leading slash. shouldn't be too hard to do.

pik94: Ok, I'll update the uri with a format `databricks://:access_token@account/http_path/catalog/schema/` in that case.

erezsh: Maybe account should be where user is? Seems to make sense.

(I don't know databricks, do they have some convention about this?)

glebmezh:
@erezsh granted access

erezsh:
@pik94  The tests are failing for numbers. They are expected to be `0.00000000...`, but instead databricks formats them as `0E-10`

For example:

```
======================================================================
test_types_mysql_num65_10_databricks_double_50 (tests.test_database_types.TestDiffCrossDatabaseTables)
  File "C:\Python310\lib\site-packages\parameterized\parameterized.py", line 533, in standalone_func
    return func(*(a + p.args), **p.kwargs)
  File "C:\code\datafold\xdiff\tests\test_database_types.py", line 630, in test_types
AssertionError: Lists differ: [] != [('-', ('1', '0.0000000000')), ('+', ('1',[116 chars]0'))]

Second list contains 6 additional elements.
First extra element 0:
('-', ('1', '0.0000000000'))

- []
+ [('-', ('1', '0.0000000000')),
+  ('+', ('1', '0E-10')),
+  ('-', ('14', '0.0000000000')),
+  ('+', ('14', '0E-10')),
+  ('-', ('15', '0.0000000000')),
+  ('+', ('15', '0E-10'))]
```

erezsh:
@pik94 I also had a look at the URI, and I don't think it's a good idea to put the "http_path" as the user, because then you'll have to escape the slashes.

I think this should be the URI instead:

```
databricks://:access_token@server_name/http_path?schema=..&catalog=..
```

This also reflects that "schema" and "catalog" are optional parameters.

I already have a patch for this, so if you agree, I'll push it on to this PR.

JCZuurmond:
:wave: Hi, y'all. Just here to say that I would like to try `data-diff` on Databricks. Happy to see this PR being merged.

PR #56, opened 2022-06-14
Quote columns; Quote snowflake schema; Force ids as ints

erezsh:
Initially I meant to make quoting the column names optional, but now I think it should always be done.

Unless I'm wrong?

### DIFF
--- a/data_diff/database.py
+++ b/data_diff/database.py
@@ -363,7 +363,7 @@ class Snowflake(Database):
         return _query_conn(self._conn, sql_code)
 
     def quote(self, s: str):
-        return s
+        return f'"{s}"'
 
     def md5_to_int(self, s: str) -> str:
         return f"BITAND(md5_number_lower64({s}), {CHECKSUM_MASK})"
@@ -371,6 +371,8 @@ class Snowflake(Database):
     def to_string(self, s: str):
         return f"cast({s} as string)"
 
+
+HELP_SNOWFLAKE_URI_FORMAT = 'snowflake://<user>:<pass>@<account>/<database>/<schema>?warehouse=<warehouse>'
 
 def connect_to_uri(db_uri: str, thread_count: Optional[int] = 1) -> Database:
     """Connect to the given database uri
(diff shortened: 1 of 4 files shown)


tests/test_diff_tables.py
```diff
@@ -123,7 +123,7 @@ def setUp(self):
 
     def test_properties_on_empty_table(self):
         self.assertEqual(0, self.table.count())
-        self.assertEqual(["id", "timestamp"], self.table._relevant_columns)
+        # self.assertEqual(["id", "timestamp"], self.table._relevant_columns)
```
sirupsen, line 126: Why is this commented out?

data_diff/diff_tables.py
```diff
@@ -231,8 +239,8 @@ def diff_tables(self, table1: TableSegment, table2: TableSegment) -> DiffResult:
         mins, maxs = zip(*key_ranges)
 
         # We add 1 because our ranges are exclusive of the end (like in Python)
-        min_key = min(mins)
-        max_key = max(maxs) + 1
+        min_key = min(map(int, mins))
+        max_key = max(map(int, maxs)) + 1
```
sirupsen, line 243: This makes sense, what did they pass?

data_diff/database.py
```diff
@@ -394,11 +397,17 @@ def connect_to_uri(db_uri: str, thread_count: Optional[int] = 1) -> Database:
...
+            schema = dsn.query['schema']
+        elif len(dsn.paths) == 2:
+            database, schema = dsn.paths
+        else:
+            raise ValueError(f"Too many parts in path. Expected format: '{HELP_SNOWFLAKE_URI_FORMAT}'")
         try:
             warehouse = dsn.query['warehouse']
         except KeyError:
-            raise ValueError("Must provide warehouse. Format: 'snowflake://<user>:<pass>@<account>/<database>/<schema>?warehouse=<warehouse>'")
+            raise ValueError(f"Must provide warehouse. Expected format: '{HELP_SNOWFLAKE_URI_FORMAT}'")
```
sirupsen, line 410: 👍🏻

sirupsen (approved):

tests/test_diff_tables.py
erezsh: Because _relevant_columns are now instances of ColumnName. Also, it's an internal object, it shouldn't be tested like this anyway.

data_diff/diff_tables.py
erezsh: Snowflake-connector returned the ids as strings.

PR #58, opened 2022-06-14
Normalize Fields (Replacing PR #54)

erezsh:
Replacing PR #54. Merged with master.

- Query database for table schema and parse it
- Always convert to normalized-repr before checksum or download
  - Support for dates with precision and timezone (needs more testing)
- Negotiate minimal mutual precision when diffing two columns (needs more testing)


data_diff/diff_tables.py
```diff
@@ -253,6 +280,28 @@ def diff_tables(self, table1: TableSegment, table2: TableSegment) -> DiffResult:
 
         return self._bisect_and_diff_tables(table1, table2)
 
+    def _validate_and_adjust_columns(self, table1, table2):
```
sirupsen, line 283: A few tests for some of this would be great in `test_diff_tables.py`

tests/test_database_types.py
```diff
@@ -0,0 +1,222 @@
...
+
+TYPE_SAMPLES = {
+    "int": [127, -3, -9, 37, 15, 127],
+    "datetime_no_timezone": [
+        "2020-01-01 15:10:10",
+        "2020-01-01 9:9:9",
+        "2022-01-01 15:10:01.139",
+        "2022-01-01 15:10:02.020409",
+        "2022-01-01 15:10:03.003030",
+        "2022-01-01 15:10:05.009900",
```
sirupsen, line 23: I think we should probably make some of these 9 precision, just for a bit of extra testing.

tests/test_database_types.py
```diff
@@ -0,0 +1,222 @@
...
+def _insert_to_table(conn, table, values):
+    insertion_query = f"INSERT INTO {table} (id, col) VALUES "
+    for j, sample in values:
+        insertion_query += f"({j}, '{sample}'),"
+
+    conn.query(insertion_query[0:-1], None)
+    if not isinstance(conn, db.BigQuery):
+        conn.query("COMMIT", None)
+
+class TestDiffCrossDatabaseTables(unittest.TestCase):
```
sirupsen, line 176: I wonder if we should skip this one by default because it's so time-consuming?

tests/test_normalize_fields.py
```diff
@@ -0,0 +1,125 @@
...
+    i = 0
+
+    def _new_table(self, name):
+        self.i += 1
+        return f"t_{self.i}"
+
+    def _test_dates_for_db(self, item, precision=6):
+        db_id, conn_string = item
+
+        print(f"Testing {db_id}")
```
sirupsen, line 35: I don't think we should print in unit tests by default

tests/test_normalize_fields.py
```diff
@@ -0,0 +1,125 @@
...
+                        pql.run_statement(f"insert into {table}(id, v) values ({index}, cast(timestamp '{date}' as {date_type}))")
+                    else:
+                        pql.run_statement(f"insert into {table}(id, v) values ({index}, timestamp '{date}')")
+            pql.commit()
+
+            conn = connect_to_uri(conn_string)
+            assert type(conn) is db_id  # Might change in the future
+            if db_id is Snowflake:
+                conn.query("alter session set timestamp_output_format = 'YYYY-MM-DD HH24:MI:SS.FF6TZH'", None)
+                conn.query("alter session set timestamp_ntz_output_format = 'YYYY-MM-DD HH24:MI:SS.FF6'", None)
```
sirupsen, line 80: Why is this necessary in this suite, if it isn't in the other one?

tests/test_normalize_fields.py
```diff
@@ -0,0 +1,125 @@
...
+import preql
+
+from data_diff.database import BigQuery, MySQL, Snowflake, connect_to_uri, Oracle, DEFAULT_PRECISION
+from data_diff.sql import Select
+from data_diff import database as db
+
+from .common import CONN_STRINGS
+
+
+DATE_TYPES = {
```
sirupsen, line 14: Do we need this test suite at all when we have the test_types one?

tests/test_normalize_fields.py
```diff
@@ -0,0 +1,125 @@
...
+            dates = [d.replace(tzinfo=None) for d in dates]
+
+        pql = preql.Preql(conn_string)
+
+        date_types = [t.format(p=precision) for t in DATE_TYPES[db_id]]
+        date_type_tables = {dt: self._new_table(dt) for dt in date_types}
+        if db_id is BigQuery:
+            date_type_tables = {dt: f"data_diff.{name}" for dt, name in date_type_tables.items()}
+        elif db_id is MySQL:
+            pql.run_statement("SET @@session.time_zone='+00:00'")
```
sirupsen, line 53: Why do you need this?

tests/test_database_types.py
sirupsen: I still get some failures, for example, here's `snowflake timestamp(9) -> postgres timestamp(6)`, you can run with `-k test_types_Snowflake_timestamp_9__to_Postgres_timestamp_6_without_time_zone`.

```
======================================================================
FAIL: test_types_Snowflake_timestamp_9__to_Postgres_timestamp_6_without_time_zone (tests.test_database_types.TestDiffCrossDatabaseTables)
  File "/Users/simon/Library/Caches/pypoetry/virtualenvs/data-diff-DzlzoNC0-py3.9/lib/python3.9/site-packages/parameterized/parameterized.py", line 533, in standalone_func
    return func(*(a + p.args), **p.kwargs)
  File "/Users/simon/src/github.com/datafold/data-diff/tests/test_database_types.py", line 236, in test_types
AssertionError: Lists differ: [] != [('-', ('6', '2022-01-01 15:10:03.003031')[305 chars]0'))]

Second list contains 8 additional elements.
First extra element 0:
('-', ('6', '2022-01-01 15:10:03.003031'))

- []
+ [('-', ('6', '2022-01-01 15:10:03.003031')),
+  ('+', ('6', '2022-01-01 15:10:03.003030')),
+  ('-', ('7', '2022-01-01 15:10:05.009901')),
+  ('+', ('7', '2022-01-01 15:10:05.009900')),
+  ('-', ('8', '2022-01-01 15:10:05.009901')),
+  ('+', ('8', '2022-01-01 15:10:05.009900')),
+  ('-', ('9', '2022-01-01 15:10:05.009901')),
+  ('+', ('9', '2022-01-01 15:10:05.009900'))]
```

Seems to be:

```
test_types_Snowflake_timestamp_9__to_MySQL_timestamp_6_
test_types_Snowflake_timestamp_9__to_Postgres_timestamp_6_with_time_zone
test_types_Snowflake_timestamp_9__to_Postgres_timestamp_6_without_time_zone
```

data_diff/diff_tables.py
```diff
@@ -61,6 +61,9 @@ class TableSegment:
     min_update: DbTime = None
     max_update: DbTime = None
 
+    quote_columns: bool = True
```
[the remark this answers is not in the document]
erezsh: We might expose it using a switch later on. It's still not clear if this is the permanent solution.

tests/test_database_types.py
erezsh: I don't know why, I don't get these failures.

tests/test_normalize_fields.py
erezsh: Just a relic. I'll remove it.

tests/test_normalize_fields.py
erezsh: It tests timezones. test_types doesn't at the moment.

tests/test_normalize_fields.py
erezsh: Because mysql might have a timezone != 0 by default, and it will mess up the tests.

tests/test_database_types.py
sirupsen: Yeah, it seems to only happen when Snowflake is the origin. Very, very weird it doesn't happen for you? Wtf?

sirupsen: `poetry run python3 -m unittest tests/test_database_types.py  -v -k test_types_Snowflake_timestamp_9__to_Postgres_timestamp_6_without_time_zone`

tests/test_normalize_fields.py
sirupsen: Ok... I think in a follow-up for time-zone support, we should remove this test suite and just have the other one only. Simplifies a lot.

erezsh: If it covers all the cases, then sure.

tests/test_database_types.py
erezsh: Sounds like lowercase/upper case shenanigans

data_diff/diff_tables.py
erezsh: `test_types` already tests it. Only the missing columns aren't tested, but we don't have a test for every exception in our code. (far from it)

tests/test_database_types.py
erezsh: Can't get it done with the way our tests work right now, using Python's Datetime to copy the values from source to target. I can't see a simple solution to it, other than hardcode this scenario.

data_diff/diff_tables.py
```diff
@@ -73,11 +76,23 @@ def __post_init__(self):
...
 
     @property
     def _update_column(self):
-        return ColumnName(self.update_column)
+        return self._quote_column(self.update_column)
+
+    def _quote_column(self, c):
+        if self.quote_columns:
+            return self.database.quote(c)
+        return c
```
sirupsen, line 88: CleanShot 2022-06-14 at 16 38 49@2x
CleanShot 2022-06-14 at 16 39 05@2x

Something's not right. I get an error pretty much no matter what I do here.

data_diff/diff_tables.py
```diff
@@ -73,11 +76,23 @@ def __post_init__(self):
...
+
+    def with_schema(self) -> "TableSegment":
```
sirupsen, line 90: For a snowflake table (see above error) this is somehow empty. I think then we'll need to error early

sirupsen: Ahhh it's because it's of a string type (I'm testing something else haha)

data_diff/diff_tables.py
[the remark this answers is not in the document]
sirupsen: I can't really get this to work locally, I think we need to do something that's friendlier to the user, or at least provide warnings that nudge people in the right direction

data_diff/diff_tables.py
erezsh: These IDs are actual strings.. We don't support that yet.

data_diff/diff_tables.py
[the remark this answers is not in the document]
erezsh: Hmm, maybe using `lower()` is a bug

PR #61, opened 2022-06-15
Added the any-case switch

erezsh:
(Based on normalized_fields PR)

### DIFF
--- a/data_diff/diff_tables.py
+++ b/data_diff/diff_tables.py
@@ -4,7 +4,7 @@
 import time
 from operator import methodcaller
 from collections import defaultdict
-from typing import List, Tuple, Iterator
+from typing import List, Tuple, Iterator, Optional
 import logging
 from concurrent.futures import ThreadPoolExecutor
 
@@ -18,6 +18,7 @@ logger = logging.getLogger("diff_tables")
 RECOMMENDED_CHECKSUM_DURATION = 10
 
 DEFAULT_BISECTION_THRESHOLD = 1024 * 16
+DEFAULT_BISECTION_FACTOR = 32
 
 
 def safezip(*args):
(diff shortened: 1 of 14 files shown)


data_diff/diff_tables.py
```diff
@@ -73,11 +96,28 @@ def __post_init__(self):
...
+        return self._quote_column(self.key_column)
 
     @property
     def _update_column(self):
-        return ColumnName(self.update_column)
+        return self._quote_column(self.update_column)
+
+    def _quote_column(self, c):
+        if self.case_sensitive:
+            return self.database.quote(c)
```
sirupsen, line 107: Instead of disabling quoting (which I think is kinda hard), why not use the case-insensitive dict to get the right name for the column?

sirupsen (commented):
I think this is basically ready to merge. Just two suggestions from some trying... mostly about Snowflake configuration 😅 

For Snowflake at least, I think `--any-case` should probably be the default, and the flag should be switched to `--case-sensitive` maybe... Now that you are warning anyway if there's a conflict.

Since tables aren't case-sensitive, can you change the error:

```python
raise RuntimeError(f"Table {'.'.join(path)} does not exist, or has no columns")
```

To say that maybe you have the casing wrong, or the schema, or database are wrong (or have the wrong casing)?

data_diff/diff_tables.py
sirupsen: As long as you always show the quoted form in the errors / exceptions, I think it's quite valid.

sirupsen: (1) Like do I think you should use the case-insensitive map to get the "correct" casing and raise an exception if there's ever multiple entries in the map to disable case-insensitivity? Yep.

(2) By "show the quoted form", I mean that in all exceptions that mentions the column, to show `"MY_primARy_KEY"` not `MY_primARy_KEY` or `my_primary_key` 👍🏻

sirupsen: Concern is especially when they have different casing between e.g. Postgres and Snowflake. Not sure what happens without this patch for that

sirupsen:
@erezsh anything I can help with to unblock this? I think we should get this in and cut a new release with your new DSN parser too asap

data_diff/diff_tables.py
erezsh: 1) That's already what we're doing..

2) Don't we already do that? 👍🏻

sirupsen: Good, wanted to confirm!

sirupsen:
I do feel that we should make this the default... especially because of casing between dbs. E.g. `id` in Postgres and `ID` in Snowflake.

erezsh:
So you think the switch should be `--case-sensitive`?

sirupsen:
@erezsh yep. Always quoting by default with the "correct" casing from the case-insensitive map

PR #62, opened 2022-06-15
updated for ssl support (presto) and role configs etc (snowflake)

sfenner-eb:


sirupsen:
Why'd you decide to close?

PR #64, opened 2022-06-16
Better parsing and nicer errors for URIs

erezsh:


data_diff/database.py
```diff
@@ -714,6 +716,65 @@ def normalize_value_by_type(self, value: str, coltype: ColType) -> str:
 HELP_PRESTO_URI_FORMAT = "presto://<user>@<host>/<catalog>/<schema>"
```
sirupsen, line 716: Consider upper-casing the SCHEMA and WAREHOUSE in the help string for Snowflake as a nudge 👍🏻

data_diff/database.py
```diff
@@ -714,6 +716,65 @@ def normalize_value_by_type(self, value: str, coltype: ColType) -> str:
...
+
+
+MATCH_URI_PATH = {
+    "postgres": MatchUriPath(Postgres, ['database?'], help_str = "postgres://<user>:<pass>@<host>/<database>"),
+    "mysql": MatchUriPath(MySQL, ['database?'], help_str = "mysql://<user>:<pass>@<host>/<database>"),
+    "oracle": MatchUriPath(Oracle, ['database?'], help_str = "oracle://<user>:<pass>@<host>/<database>"),
+    "mssql": MatchUriPath(MsSQL, ['database?'], help_str = "mssql://<user>:<pass>@<host>/<database>"),
+    "redshift": MatchUriPath(Redshift, ['database?'], help_str = "redshift://<user>:<pass>@<host>/<database>"),
+    "snowflake": MatchUriPath(Snowflake, ['database', 'schema'], ['warehouse'], help_str = HELP_SNOWFLAKE_URI_FORMAT),
+    "presto": MatchUriPath(Presto, ['catalog', 'schema'], help_str = HELP_PRESTO_URI_FORMAT),
```
sirupsen, line 773: Why are some constants, and some inlined?

data_diff/database.py
```diff
@@ -714,6 +716,65 @@ def normalize_value_by_type(self, value: str, coltype: ColType) -> str:
 HELP_PRESTO_URI_FORMAT = "presto://<user>@<host>/<catalog>/<schema>"
 
 
+@dataclass
+class MatchUriPath:
```
sirupsen, line 720: I think a couple of quick unit tests for this class would be helpful

data_diff/database.py
```diff
@@ -714,6 +716,65 @@ def normalize_value_by_type(self, value: str, coltype: ColType) -> str:
...
+    database_cls: type
+    params: List[str]
+    kwparams: List[str] = []
+    help_str: str
+
+    def match_path(self, dsn):
+        dsn_dict = dict(dsn.query)
+        matches = {}
+        for param, arg in zip_longest(self.params, dsn.paths):
+            if param is None:
```
sirupsen, line 730: What is it's empty, or does that become None in the DSN parser?

data_diff/database.py
```diff
@@ -726,63 +787,35 @@ def connect_to_uri(db_uri: str, thread_count: Optional[int] = 1) -> Database:
...
+        raise NotImplementedError(f"Scheme {scheme} currently not supported")
+
+    cls = matcher.database_cls
+    kw = matcher.match_path(dsn)
+
+    if scheme == "bigquery":
+        return cls(dsn.host, **kw)
+
+    if issubclass(cls, ThreadedDatabase):
+        return cls(dsn.host, dsn.port, dsn.user, dsn.password, thread_count=thread_count, **kw)
```
sirupsen, line 818: Is there a way we can just pass arbitrary key-word arguments? For example, say I want to connect to Postgres with `sslmode='require'`, I would expect to be able to pass that as part of the URI

sirupsen (approved):
Tried this with a bunch of bad connection strings, and it's quite nice! Well done!

data_diff/database.py
erezsh: Good idea

data_diff/database.py
erezsh: No reason. I can make them all inlined.

data_diff/database.py
erezsh: What's empty?

sirupsen: If I pass `//database`

data_diff/database.py
erezsh: That's what **kw is for.

sirupsen: Right, should've been more specific, for e.g. the Postgres driver they're not passed along:

```
postgres://postgres:Password1@localhost/postgres?sslmode=require
```

Will give `TypeError: __init__() got an unexpected keyword argument 'sslmode'`

erezsh: Oh, you mean pass everything directly to the postgres driver.. I guess we can do that. I was worried it would might cause confusion or strange behaviors, though I can't say exactly how or why.

sirupsen: We can wait until it arises as an issue if you like. I just think that'd be what I expected :)

data_diff/database.py
sirupsen: kk

data_diff/database.py
erezsh: It's fine, I don't see any harm in it. Added.

data_diff/database.py
erezsh: Done

PR #66, opened 2022-06-16
<No Merge> Data diff fork compare

sfenner-eb:
Comparing across forks, not to merge

### DIFF
--- a/data_diff/database.py
+++ b/data_diff/database.py
@@ -65,7 +65,6 @@ def _one(seq):
 
 def _query_conn(conn, sql_code: str) -> list:
     c = conn.cursor()
-    sleep(.5)
     c.execute(sql_code)
     return c.fetchall()
 
@@ -146,6 +145,7 @@ class ThreadedDatabase(Database):
         self.thread_local.conn = self.create_connection()
 
     def _query(self, sql_code: str):
+        # print(sql_code)
         r = self._queue.submit(self._query_in_worker, sql_code)
         return r.result()
 
@@ -204,6 +204,7 @@ class Presto(Database):
                 schema=schema,
                 catalog=database,
                 http_scheme='https',
+                http_headers={'X-Presto-Time-Zone': 'UTC'},
                 source="odbc",
                 auth=prestodb.auth.BasicAuthentication(user, password),
             )
@@ -225,7 +226,7 @@ class Presto(Database):
 
     def _query(self, sql_code: str) -> list:
         "Uses the standard SQL cursor interface"
-        print(sql_code)
+        # print(sql_code)
         return _query_conn(self._conn, sql_code)
 
 
@@ -401,7 +402,7 @@ class Snowflake(Database):
         return f"BITAND(md5_number_lower64({s}), {CHECKSUM_MASK})"
 
     def to_string(self, s: str):
-        return f"cast({s} as string)"
+        return f"cast({s} as varchar)"
 
 
 HELP_SNOWFLAKE_URI_FORMAT = 'snowflake://<user>:<pass>@<account>/<database>/<schema>?warehouse=<warehouse>'
--- a/data_diff/database.py
+++ b/data_diff/database.py
@@ -4,8 +4,11 @@ from typing import Tuple, Optional
 from concurrent.futures import ThreadPoolExecutor
 import threading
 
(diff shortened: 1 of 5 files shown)


data_diff/database.py
```diff
@@ -188,11 +192,28 @@ def to_string(self, s: str):
...
+        def get_presto_connection(user, password):
+            import prestodb
+            conn = prestodb.dbapi.connect(
+                host=host,
+                port=port,
+                user=user,
+                schema=schema,
+                catalog=database,
+                http_scheme='https',
+                http_headers={'X-Presto-Time-Zone': 'UTC'},
```
joeb-eb, line 207: can take this out......

data_diff/database.py
```diff
@@ -188,11 +190,27 @@ def to_string(self, s: str):
 
 
 class Presto(Database):
-    def __init__(self, host, port, database, user, password):
+    def __init__(self, host, port, user, password, schema, database, print_sql=True):
```
joeb-eb, line 193: maybe we can update this signature to pass in "extra" args, then pass those extra args to `prestodb.dbapi.connect` might be a lighter weight change.

data_diff/database.py
```diff
@@ -337,12 +355,22 @@ def __init__(
...
         schema: str,
         database: str,
-        role: str = None,
+        role: str,
         print_sql: bool = False,
     ):
         snowflake = import_snowflake()
         logging.getLogger("snowflake.connector").setLevel(logging.WARNING)
-
+        with open("/Users/sfenner/.ssh/snowflake_rsa_key.p8", "rb") as key:
```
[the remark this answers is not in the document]
joeb-eb: perhaps there is a more generic way to do this type of thing......

data_diff/database.py
```diff
@@ -408,7 +436,11 @@ def connect_to_uri(db_uri: str, thread_count: Optional[int] = 1) -> Database:
             warehouse = dsn.query['warehouse']
         except KeyError:
             raise ValueError(f"Must provide warehouse. Expected format: '{HELP_SNOWFLAKE_URI_FORMAT}'")
-        return Snowflake(dsn.host, dsn.user, dsn.password, warehouse=warehouse, database=database, schema=schema)
+        try:
+            role = dsn.query['role']
+        except KeyError:
+            raise ValueError(f"Must provide role. Expected format: '{HELP_SNOWFLAKE_URI_FORMAT}'")
```
joeb-eb, line 442: Does the Snowflake constructor throw a similar error? I assume it is ok with out one for users that have a default role assigned.

That isn't true in our case and we'd like to pass that along, but maybe isn't quite the same as the other objects like warehouse, database, and scheme....... Not sure.

data_diff/diff_tables.py
```diff
@@ -10,7 +10,7 @@
 
 from runtype import dataclass
 
-from .sql import Select, Checksum, Compare, DbPath, DbKey, DbTime, Count, TableName, Time, Min, Max, ColumnName
+from .sql import Select, Checksum, Compare, DbPath, DbKey, DbTime, Count, TableName, Time, Min, Max, ColumnName, Compiler
```
joeb-eb, line 13: I don't see `Compiler` used in this file. We should take this out.

data_diff/sql.py
```diff
@@ -113,8 +113,13 @@ class Checksum(Sql):
     exprs: List[SqlOrStr]
 
     def compile(self, c: Compiler):
-        compiled_exprs = ", ".join(c.database.to_string(s) for s in map(c.compile, self.exprs))
-        expr = f"concat({compiled_exprs})"
+        compiled_exprs_raw = [c.database.to_string(s) for s in map(c.compile, self.exprs)]
```
joeb-eb, line 116: Could you provide an example here or in the description of what prompted this change?

data_diff/database.py
sirupsen: I think excess kw should just be passed straight to the driver

data_diff/database.py
erezsh: This code is out-dated

data_diff/database.py
erezsh: This code is out-dated

PR #67, opened 2022-06-17
tests: database_types dual-use for benchmarks

sirupsen:
@erezsh still having some trouble with Presto, but the idea is to dual-use this test-suite for benchmarking too by upping `N_SAMPLES`.

### DIFF
--- a/data_diff/diff_tables.py
+++ b/data_diff/diff_tables.py
@@ -378,6 +378,12 @@ class TableDiffer:
             f"size: {table2.max_key-table1.min_key}"
         )
 
+        # The entire segment wasn't below the threshold, but the next set of
+        # segments might be. In that case, it's useless to checksum them.
+        max_rows_from_keys = max(table1.max_key - table1.min_key, table2.max_key - table2.min_key)
+        if max_rows_from_keys < self.bisection_threshold:
+            yield from self._bisect_and_diff_tables(table1, table2, level=level, max_rows=max_rows_from_keys)
+
         (count1, checksum1), (count2, checksum2) = self._threaded_call("count_and_checksum", [table1, table2])
 
         if count1 == 0 and count2 == 0:
--- a/tests/test_database_types.py
+++ b/tests/test_database_types.py
@@ -1,30 +1,88 @@
 import unittest
 import preql
 import time
+import datetime
 from data_diff import database as db
 from data_diff.diff_tables import TableDiffer, TableSegment, split_space
 from parameterized import parameterized, parameterized_class
-from .common import CONN_STRINGS, str_to_checksum
+from .common import CONN_STRINGS, DEFAULT_N_SAMPLES, N_SAMPLES, str_to_checksum
 import logging
 
-logging.getLogger("diff_tables").setLevel(logging.WARN)
-logging.getLogger("database").setLevel(logging.WARN)
-
 CONNS = {k: db.connect_to_uri(v) for k, v in CONN_STRINGS.items()}
-
 CONNS[db.MySQL].query("SET @@session.time_zone='+00:00'", None)
 
+
+class Faker:
+    pass
+
+
+class DateTimeFaker:
+    def __init__(self, max):
+        self.max = max
+
+    def __iter__(self):
+        self.prev = datetime.datetime(2000, 1, 1, 0, 0, 0, 0)
+        self.i = 0
+        return self
+
+    def __len__(self):
+        return self.max
+
+    def __next__(self) -> str:
+        if self.i < self.max:
+            self.prev = self.prev + datetime.timedelta(seconds=3, microseconds=571)
+            self.i += 1
+            return str(self.prev)
+        else:
+            raise StopIteration
+
+
+class IntFaker:
+    def __init__(self, max):
+        self.max = max
+
+    def __iter__(self):
+        self.prev = -128
+        self.i = 0
+        return self
+
+    def __len__(self):
+        return self.max
+
+    def __next__(self) -> str:
+        if self.i < self.max:
+            self.prev += 1
+            self.i += 1
+            return str(self.prev)
+        else:
+            raise StopIteration
+
+
+class FloatFaker:
+    def __init__(self, max):
+        self.max = max
+
+    def __iter__(self):
+        self.prev = -10.0001
+        self.i = 0
+        return self
+
+    def __len__(self):
+        return self.max
+
+    def __next__(self) -> str:
+        if self.i < self.max:
+            self.prev += 0.00571
+            self.i += 1
+            return str(self.prev)
+        else:
+            raise StopIteration
+
+
 TYPE_SAMPLES = {
-    "int": [127, -3, -9, 37, 15, 127],
-    "datetime_no_timezone": [
-        "2020-01-01 15:10:10",
-        "2020-01-01 9:9:9",
-        "2022-01-01 15:10:01.139",
-        "2022-01-01 15:10:02.020409",
-        "2022-01-01 15:10:03.003030",
-        "2022-01-01 15:10:05.009900",
-    ],
-    "float": [0.0, 0.1, 0.10, 10.0, 100.98],
+    "int": IntFaker(N_SAMPLES),
+    "datetime_no_timezone": DateTimeFaker(N_SAMPLES),
+    "float": FloatFaker(N_SAMPLES),
 }
 
 DATABASE_TYPES = {
@@ -150,9 +208,12 @@ type_pairs = []
 # target_type: (int, bigint) }
 for source_db, source_type_categories in DATABASE_TYPES.items():
     for target_db, target_type_categories in DATABASE_TYPES.items():
-        for type_category, source_types in source_type_categories.items():  # int, datetime, ..
+        for (
+            type_category,
+            source_types,
+        ) in source_type_categories.items():  # int, datetime, ..
             for source_type in source_types:
-                for target_type in target_type_categories[type_category]:
+                for target_type in target_type_categories.get(type_category, []):
                     if CONNS.get(source_db, False) and CONNS.get(target_db, False):
                         type_pairs.append(
                             (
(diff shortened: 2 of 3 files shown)


data_diff/database.py
```diff
@@ -381,7 +381,13 @@ def to_string(self, s: str):
 
     def _query(self, sql_code: str) -> list:
         "Uses the standard SQL cursor interface"
-        return _query_conn(self._conn, sql_code)
+        c = self._conn.cursor()
```
erezsh, line 384: If this is Presto-specific, it should only be used for Presto

data_diff/diff_tables.py
```diff
@@ -378,6 +378,13 @@ def _diff_tables(self, table1, table2, level=0, segment_index=None, segment_coun
             f"size: {table2.max_key-table1.min_key}"
         )
 
+        # The entire segment wasn't below the threshold, but the next set of
```
erezsh, line 381: This belongs in a separate PR

tests/test_database_types.py
```diff
@@ -1,30 +1,88 @@
...
-    "datetime_no_timezone": [
-        "2020-01-01 15:10:10",
-        "2020-01-01 9:9:9",
-        "2022-01-01 15:10:01.139",
-        "2022-01-01 15:10:02.020409",
-        "2022-01-01 15:10:03.003030",
-        "2022-01-01 15:10:05.009900",
-    ],
-    "float": [0.0, 0.1, 0.10, 10.0, 100.98],
+    "int": IntFaker(N_SAMPLES),
```
erezsh, line 83: This feels like a step backwards. How can we be sure that the "faker" creates all the edge cases that we are interested in?

tests/test_database_types.py
```diff
@@ -150,9 +211,12 @@
...
 for source_db, source_type_categories in DATABASE_TYPES.items():
     for target_db, target_type_categories in DATABASE_TYPES.items():
-        for type_category, source_types in source_type_categories.items():  # int, datetime, ..
+        for (
+            type_category,
+            source_types,
+        ) in source_type_categories.items():  # int, datetime, ..
             for source_type in source_types:
-                for target_type in target_type_categories[type_category]:
+                for target_type in target_type_categories.get(type_category, []):
```
erezsh, line 219: We'll never know if a typo caused some the tests to not run.

tests/test_database_types.py
```diff
@@ -178,60 +242,135 @@ def expand_params(testcase_func, param_num, param):
...
         src_table = src_conn.quote(".".join(src_table_path))
+
+        dst_table_name = f"dst_{target_db.__name__.lower()}_{parameterized.to_safe_name(target_type)}_{src_table_name}"
+        # Since we insert `src` to `dst`, the `dst` can be different for the
+        # same type. 😭
+        dst_table_path = dst_conn.parse_table_name(dst_table_name)
         dst_table = dst_conn.quote(".".join(dst_table_path))
 
+        # If you want clean house from e.g. changing fakers.
         src_conn.query(f"DROP TABLE IF EXISTS {src_table}", None)
```
erezsh, line 271: Oracle doesn't support "DROP TABLE IF EXISTS"

tests/test_database_types.py
```diff
@@ -178,60 +242,135 @@ def expand_params(testcase_func, param_num, param):
...
-        dst_conn.query(f"DROP TABLE IF EXISTS {dst_table}", None)
-        dst_conn.query(f"CREATE TABLE {dst_table}(id int, col {target_type});", None)
-        _insert_to_table(dst_conn, dst_table, values_in_source)
+        start = time.time()
+        already_seeded = self._create_table(src_conn, src_table, source_type)
+        if not already_seeded:
+            self._insert_to_table(
+                src_conn, src_table, source_type, enumerate(sample_values, 1)
+            )
+        insertion_source_duration = round(time.time() - start, 2)
```
erezsh, line 287: Why 2? Needs a comment

tests/test_database_types.py
```diff
@@ -178,60 +242,135 @@ def expand_params(testcase_func, param_num, param):
...
 
-        self.table = TableSegment(self.src_conn, src_table_path, "id", None, ("col",), quote_columns=False)
-        self.table2 = TableSegment(self.dst_conn, dst_table_path, "id", None, ("col",), quote_columns=False)
+        start = time.time()
+        already_seeded = self._create_table(dst_conn, dst_table, target_type)
```
erezsh, line 290: We should be dropping the tables when we're done, no? So this exists only in case the tests fail.. And what if it's an old table with different values? Better to drop the tables and re-create everything.

tests/test_database_types.py
```diff
@@ -178,60 +242,135 @@ def expand_params(testcase_func, param_num, param):
...
+        start = time.time()
         self.assertEqual(len(sample_values), self.table.count())
+        count1_duration = round(time.time() - start, 2)
+
+        start = time.time()
         self.assertEqual(len(sample_values), self.table2.count())
+        count2_duration = round(time.time() - start, 2)
+
+        # Ensure we actually checksum even for few samples!
+        ch_threshold = 5_000 if N_SAMPLES > DEFAULT_N_SAMPLES else 3
```
erezsh, line 305: This needs an explanation

tests/test_database_types.py
```diff
@@ -178,60 +242,135 @@ def expand_params(testcase_func, param_num, param):
...
+            bisection_factor=dl_factor,
+            max_threadpool_size=dl_threads,
+        )
         diff = list(differ.diff_tables(self.table, self.table2))
         expected = []
         self.assertEqual(expected, diff)
         self.assertEqual(len(sample_values), differ.stats.get("rows_downloaded", 0))
+        download_duration = round(time.time() - start, 2)
+
+        print(
```
erezsh, line 343: I thought we agreed tests don't print. Use the logger.

tests/test_database_types.py
```diff
@@ -178,60 +242,135 @@ def expand_params(testcase_func, param_num, param):
...
+            f"\nsource_db={source_db.__name__} target_db={target_db.__name__} rows={N_SAMPLES} src_table={src_table} target_table={dst_table} source_type={repr(source_type)} target_type={repr(target_type)} insertion_source={insertion_source_duration}s insertion_target={insertion_target_duration}s count_source={count1_duration}s count_target={count2_duration}s checksum={checksum_duration}s download={download_duration}s download_threads={dl_threads} download_bisection_factor={dl_factor} download_bisection_threshold={dl_threshold} checksum_threads={ch_threads} checksum_bisection_factor={ch_factor} checksum_bisection_threshold={ch_threshold}"
+        )
+
+    def _create_table(self, conn, table, type) -> bool:
+        conn.query(f"CREATE TABLE IF NOT EXISTS {table}(id int, col {type})", None)
```
erezsh, line 348: Not all dialects support IF NOT EXISTS here.

tests/test_database_types.py
```diff
@@ -178,60 +242,135 @@ def expand_params(testcase_func, param_num, param):
...
+        # print(conn.query("SHOW TABLES FROM public", list))
+        # print(conn.query("SELECT * FROM INFORMATION_SCHEMA.TABLES", list))
+        already_seeded = conn.query(f"SELECT COUNT(*) FROM {table}", int) == N_SAMPLES
+        return already_seeded
+
+    def _insert_to_table(self, conn, table, type, values):
+        conn.query(f"TRUNCATE {table}", None)
```
erezsh, line 355: Did you make sure TRUNCATE exists in all dialects?

erezsh (changes requested):

data_diff/database.py
sirupsen: I'm not sure I understand what you mean 👀  This is on the Presto driver

data_diff/diff_tables.py
sirupsen: Sure. The reason I included it here is because this is how the 'download' is a fair benchmark comparison where it splits into multiple segments, processing those in threads

tests/test_database_types.py
sirupsen: I think it generates plenty of edge-cases naturally by creating lots of 9s, 8s, etc.

```
['2000-01-01 00:00:03.000571', '2000-01-01 00:00:06.001142', '2000-01-01 00:00:09.001713', '2000-01-01 00:00:12.002284', '2000-01-01 00:00:15.002855', '2000-01-01 00:00:18.003426', '2000-01-01 00:00:21.003997', '2000-01-01 00:00:24.004568', '2000-01-01 00:00:27.005139', '2000-01-01 00:00:30.005710', '2000-01-01 00:00:33.006281', '2000-01-01 00:00:36.006852', '2000-01-01 00:00:39.007423', '2000-01-01 00:00:42.007994', '2000-01-01 00:00:45.008565', '2000-01-01 00:00:48.009136', '2000-01-01 00:00:51.009707', '2000-01-01 00:00:54.010278', '2000-01-01 00:00:57.010849', '2000-01-01 00:01:00.011420', '2000-01-01 00:01:03.011991', '2000-01-01 00:01:06.012562', '2000-01-01 00:01:09.013133', '2000-01-01 00:01:12.013704', '2000-01-01 00:01:15.014275', '2000-01-01 00:01:18.014846', '2000-01-01 00:01:21.015417', '2000-01-01 00:01:24.015988', '2000-01-01 00:01:27.016559', '2000-01-01 00:01:30.017130', '2000-01-01 00:01:33.017701', '2000-01-01 00:01:36.018272', '2000-01-01 00:01:39.018843', '2000-01-01 00:01:42.019414', '2000-01-01 00:01:45.019985', '2000-01-01 00:01:48.020556', '2000-01-01 00:01:51.021127', '2000-01-01 00:01:54.021698', '2000-01-01 00:01:57.022269', '2000-01-01 00:02:00.022840', '2000-01-01 00:02:03.023411', '2000-01-01 00:02:06.023982', '2000-01-01 00:02:09.024553', '2000-01-01 00:02:12.025124', '2000-01-01 00:02:15.025695', '2000-01-01 00:02:18.026266', '2000-01-01 00:02:21.026837', '2000-01-01 00:02:24.027408', '2000-01-01 00:02:27.027979', '2000-01-01 00:02:30.028550']
```

tests/test_database_types.py
sirupsen: Fair! Reason I had it was to just be able to comment out something quickly, but I see now that the risk isn't worth it. Agree

tests/test_database_types.py
sirupsen: I'm actually going to remove/comment out these lines. The whole idea is to keep these tables around so we can re-run benchmarks without building up millions of rows.

tests/test_database_types.py
sirupsen: To avoid having `0.02387912019010s` in the logs

tests/test_database_types.py
sirupsen: We keep the tables around on purpose for performance with large `N_SAMPLES`. Only if you change the fakers would you need to drop them.

tests/test_database_types.py
sirupsen: Yep, for sure, just a debugging relic! Will log to `debug`!

tests/test_database_types.py
sirupsen: Ah, is it Oracle? I've run this on all but Redshift + Oracle.

tests/test_database_types.py
sirupsen: It's worked on all I've seen so far. It's fairly standard SQL... I mean until we discover it's not.

tests/test_database_types.py
erezsh: But there are edge-cases we already know are problematic. It doesn't make sense to throw them away, and hope we get them through an equation..

tests/test_database_types.py
erezsh: You can do that while printing or formatting. 

```python
>>> x = 0.02387912019010
>>> print(f'{x:.2f}')
0.02
```

tests/test_database_types.py
erezsh: But that's exactly what's going to happen if I check-out some old revision on git. And I won't understand why the tests are behaving strangely.

tests/test_database_types.py
erezsh: I think Oracle, yes.

tests/test_database_types.py
sirupsen: Ok. I can hard-code the first N of the faker to the existing hard-coded values

data_diff/database.py
erezsh: Okay, my bad.

tests/test_database_types.py
sirupsen: OK, that's fair. I'm thinking then I will add a benchmarking "mode" via an environment variable where you pass the sample numbers. We need to select a small subset of the tests for this either way. I'll comment it for now to get the numbers I need.

data_diff/database.py
```diff
@@ -381,7 +381,13 @@ def to_string(self, s: str):
...
+        c.execute(sql_code)
+        if sql_code.lower().startswith("select"):
+            return c.fetchall()
+        # Required for the query to actually run 🤯
+        if re.match(r"\A(insert|create)", sql_code, re.IGNORECASE):
```
erezsh, line 389: \A is not necessary. Python's `match()` doesn't scan the string.

tests/test_database_types.py
erezsh: That's acceptable.

PR #69, opened 2022-06-17
README: fix error in Python import statement

jaypeedevlin:
I had to adjust this to get the the import to work.


erezsh:
Thanks!

PR #70, opened 2022-06-17
Fixed oracle & redshift support (normalize-fields)

erezsh:


tests/test_database_types.py
```diff
@@ -204,14 +212,22 @@ def test_types(self, source_db, target_db, source_type, target_type, type_catego
         src_table = src_conn.quote(".".join(src_table_path))
         dst_table = dst_conn.quote(".".join(dst_table_path))
 
-        src_conn.query(f"DROP TABLE IF EXISTS {src_table}", None)
-        src_conn.query(f"CREATE TABLE {src_table}(id int, col {source_type});", None)
+        if isinstance(src_conn, db.Oracle):
+            src_conn.query(f"DROP TABLE {src_table}", None)
```
sirupsen, line 216: Don't we need to rescue an exception here if it doesn't support `IF EXISTS`?

tests/test_database_types.py
```diff
@@ -179,11 +178,20 @@ def expand_params(testcase_func, param_num, param):
...
-        insertion_query += f"({j}, '{sample}'),"
-
-    conn.query(insertion_query[0:-1], None)
+    insertion_query = f"INSERT INTO {table} (id, col) "
+
+    if isinstance(conn, db.Oracle):
+        selects = []
+        for j, sample in values:
+            selects.append( f"SELECT {j}, timestamp '{sample}' FROM dual" )
+        insertion_query += ' UNION ALL '.join(selects)
```
sirupsen, line 187: Wait, what? Oracle doesn't support values?

data_diff/database.py
```diff
@@ -533,6 +534,21 @@ class Redshift(Postgres):
...
+            if coltype.rounds:
+                timestamp = f"{value}::timestamp(6)"
+                secs = f"timestamp 'epoch' + round(extract(epoch from {timestamp})::decimal(38)"
+                ms = f"extract(ms from {timestamp})"
+                us = f"extract(us from {timestamp})"
+                epoch = f"{secs}*1000000 + {ms}*1000 + {us}"
+                timestamp6 = f"to_char({epoch}, -6+{coltype.precision}) * interval '0.000001 seconds', 'YYYY-mm-dd HH24:MI:SS.US')"
+            else:
+                timestamp6 = f"to_char({value}::timestamp(6), 'YYYY-mm-dd HH24:MI:SS.US')"
+            return f"RPAD(LEFT({timestamp6}, {TIMESTAMP_PRECISION_POS+coltype.precision}), {TIMESTAMP_PRECISION_POS+6}, '0')"
```
sirupsen, line 548: Can you add some more comments on the madness you went through here? Lots of casting and calculations to make this work

data_diff/database.py
```diff
@@ -509,9 +510,7 @@ def select_table_schema(self, path: DbPath) -> str:
 
     def normalize_value_by_type(self, value: str, coltype: ColType) -> str:
         if isinstance(coltype, PrecisionType):
-            if coltype.precision == 0:
-                return f"to_char(cast({value} as timestamp({coltype.precision})), 'YYYY-MM-DD HH24:MI:SS')"
-            return f"to_char(cast({value} as timestamp({coltype.precision})), 'YYYY-MM-DD HH24:MI:SS.FF{coltype.precision or ''}')"
+            return f"to_char(cast({value} as timestamp({coltype.precision})), 'YYYY-MM-DD HH24:MI:SS.FF6')"
```
sirupsen, line 513: Why is this always 6?

erezsh: That's the format I settled on. Always show 6 digits, no matter the precision.

data_diff/database.py
erezsh: Okay.

tests/test_database_types.py
erezsh: Only Oracle doesn't support if exists.

sirupsen: Yes, but what about on the first run when the table doesn't exist?

tests/test_database_types.py
erezsh: I think in Oracle VALUES can only specify one row.

tests/test_database_types.py
erezsh: Fixed.

tests/test_database_types.py
sirupsen: 😢 all good, Oracle is way more of an outlier than I thought

sirupsen (approved):

sirupsen:
Can you release a new version to Pip when you merge? 👍🏻

PR #71, opened 2022-06-17
Now automatically fixing the column case using the schema.

erezsh:
Added the `--keep-column-case` switch to disable it.

This is proposed instead of PR #61


sirupsen (approved):
LGTM, but can't test it because of a Snowflake issue I mentioned in Slack

I am thinking this will fix the issues where e.g. you pass `-k ID` and it will now work for Postgres <-> Snowflake (and `-k id`, the default, will now work too). I assume this is an issue, but I can't test right now for some reason

sirupsen:
CleanShot 2022-06-17 at 13 10 58@2x

Tested, this is the scenario this should prevent

PR #74, opened 2022-06-20
Adds support for Numeric types with arbitrary precision

erezsh:
Tested on all supported DBs.

We still need more thorough tests, to cover all db types, and more edge-case values.

### DIFF
--- a/data_diff/database.py
+++ b/data_diff/database.py
@@ -816,10 +816,12 @@ class BigQuery(Database):
 
                 timestamp6 = f"FORMAT_TIMESTAMP('%F %H:%M:%E6S', {value})"
                 return f"RPAD(LEFT({timestamp6}, {TIMESTAMP_PRECISION_POS+coltype.precision}), {TIMESTAMP_PRECISION_POS+6}, '0')"
+        elif isinstance(coltype, Integer):
+            pass
 
         elif isinstance(coltype, NumericType):
             # value = f"cast({value} as decimal)"
-            return f"format('%.{coltype.precision}f', cast({value} as decimal))"
+            return f"format('%.{coltype.precision}f', ({value}))"
 
         return self.to_string(f"{value}")
 
--- a/data_diff/database.py
+++ b/data_diff/database.py
@@ -109,6 +109,7 @@ class Datetime(TemporalType):
 
 @dataclass
 class NumericType(ColType):
+    # 'precision' signifies how many fractional digits (after the dot) we want to compare
     precision: int
 
 
@@ -183,6 +184,19 @@ class AbstractDatabase(ABC):
 
             Rounded up/down according to coltype.rounds
 
+        - Floats/Decimals are expected in the format
+            "I.P"
+
+            Where I is the integer part of the number (as many digits as necessary),
+            and must be at least one digit (0).
+            P is the fractional digits, the amount of which is specified with
+            coltype.precision. Trailing zeroes may be necessary.
+
+            Note: This precision is different than the one used by databases. For decimals,
+            it's the same as "numeric_scale", and for floats, who use binary precision,
+            it can be calculated as log10(2**p)
+
+
         """
         ...
 
@@ -234,6 +248,7 @@ class Database(AbstractDatabase):
         self._interactive = True
 
     def _convert_db_precision_to_digits(self, p: int) -> int:
+        """Convert from binary precision, used by floats, to decimal precision."""
         # See: https://en.wikipedia.org/wiki/Single-precision_floating-point_format
         return math.floor(math.log(2**p, 10))
 
--- a/data_diff/database.py
+++ b/data_diff/database.py
@@ -587,16 +587,24 @@ class Oracle(ThreadedDatabase):
         (table,) = path
 
         return (
-            f"SELECT column_name, data_type, 6 as datetime_precision, data_precision as numeric_precision"
+            f"SELECT column_name, data_type, 6 as datetime_precision, data_precision as numeric_precision, data_scale as numeric_scale"
             f" FROM USER_TAB_COLUMNS WHERE table_name = '{table.upper()}'"
         )
 
     def normalize_value_by_type(self, value: str, coltype: ColType) -> str:
         if isinstance(coltype, TemporalType):
             return f"to_char(cast({value} as timestamp({coltype.precision})), 'YYYY-MM-DD HH24:MI:SS.FF6')"
+        elif isinstance(coltype, NumericType):
+            # FM999.9990
+            format_str = "FM" + "9" * (38 - coltype.precision)
+            if coltype.precision:
+                format_str += "0." + "9" * (coltype.precision - 1) + "0"
+            return f"to_char({value}, '{format_str}')"
         return self.to_string(f"{value}")
 
-    def _parse_type(self, type_repr: str, datetime_precision: int = None, numeric_precision: int = None) -> ColType:
+    def _parse_type(
+        self, type_repr: str, datetime_precision: int = None, numeric_precision: int = None, numeric_scale: int = None
+    ) -> ColType:
         """ """
         regexps = {
             r"TIMESTAMP\((\d)\) WITH LOCAL TIME ZONE": Timestamp,
@@ -610,6 +618,18 @@ class Oracle(ThreadedDatabase):
                     precision=datetime_precision if datetime_precision is not None else DEFAULT_DATETIME_PRECISION,
                     rounds=self.ROUNDS_ON_PREC_LOSS,
                 )
+
+        cls = {
+            "NUMBER": Decimal,
+            "FLOAT": Float,
+        }.get(type_repr, None)
+        if cls:
+            if issubclass(cls, Decimal):
+                assert numeric_scale is not None, (type_repr, numeric_precision, numeric_scale)
+                return cls(precision=numeric_scale)
+
+            assert issubclass(cls, Float)
+            return cls(precision=(numeric_precision if numeric_precision is not None else 15) // 3)
 
         return UnknownColType(type_repr)
 
--- a/data_diff/database.py
+++ b/data_diff/database.py
@@ -239,13 +239,14 @@ class Database(AbstractDatabase):
 
         cls = self.NUMERIC_TYPES.get(type_repr)
         if cls:
-            assert numeric_precision is not None
             if cls is Decimal:
-                assert numeric_scale is not None
+                assert numeric_precision is not None
+                assert numeric_scale is not None, (type_repr, numeric_precision, numeric_scale)
                 return cls(precision=numeric_scale)
 
-            assert numeric_scale is None
-            return cls(precision=numeric_precision // 3)
+            assert cls is Float
+            # assert numeric_scale is None
+            return cls(precision=(numeric_precision if numeric_precision is not None else 15) // 3)
 
         return UnknownColType(type_repr)
 
@@ -642,6 +643,13 @@ class BigQuery(Database):
         "TIMESTAMP": Timestamp,
         "DATETIME": Datetime,
     }
+    NUMERIC_TYPES = {
+        "INT64": Decimal,
+        "INT32": Decimal,
+        "NUMERIC": Decimal,
+        "FLOAT64": Float,
+        "FLOAT32": Float,
+    }
     ROUNDS_ON_PREC_LOSS = False  # Technically BigQuery doesn't allow implicit rounding or truncation
 
     def __init__(self, project, *, dataset, **kw):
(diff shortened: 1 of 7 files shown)


sirupsen:
Would you be able to supply a description of what you're doing here to assist with my review tomorrow? You can either do this by updating the code with comments, or leaving a comment here

I'm finding it fairly hard to follow (and thus review), since we haven't talked as much about this as we did for datetime

erezsh:
Added some comments, hope it helps. If not, find me in the chat and I'll explain it live.

tests/test_database_types.py
```diff
@@ -37,15 +44,15 @@
         ],
         # https://www.postgresql.org/docs/current/datatype-datetime.html
         "datetime_no_timezone": [
-            "timestamp(6) without time zone",
-            "timestamp(3) without time zone",
-            "timestamp(0) without time zone",
+            # "timestamp(6) without time zone",
+            # "timestamp(3) without time zone",
+            # "timestamp(0) without time zone",
```
sirupsen, line 49: Before you merge, make sure to uncomment these again 👍🏻

tests/test_database_types.py
```diff
@@ -184,25 +212,37 @@ def _insert_to_table(conn, table, values):
...
     conn.query(insertion_query, None)
     if not isinstance(conn, db.BigQuery):
         conn.query("COMMIT", None)
 
+
 def _drop_table_if_exists(conn, table):
     with suppress(db.QueryError):
         if isinstance(conn, db.Oracle):
             conn.query(f"DROP TABLE {table}", None)
+            conn.query(f"DROP TABLE {table}", None)
```
sirupsen, line 241: Why'd you need this twice?

data_diff/diff_tables.py
```diff
@@ -364,11 +364,23 @@ def _validate_and_adjust_columns(self, table1, table2):
...
                 table2._schema[c] = col2.replace(precision=lowest.precision, rounds=lowest.rounds)
 
+            elif isinstance(col1, NumericType):
+                if not isinstance(col2, NumericType):
+                    raise TypeError(f"Incompatible types for column {c}:  {col1} <-> {col2}")
+
+                lowest = min(col1, col2, key=attrgetter("precision"))
+
+                if col1.precision != col2.precision:
+                    logger.warning(f"Using reduced precision {lowest} for column '{c}'. Types={col1}, {col2}")
```
sirupsen, line 379: For the types, it might be nice to print the database type as well with the type. Not critical right now, but I think we need to get in that habit... Seen so many errors where I'm wondering whether it's from the left or right table 😅

data_diff/diff_tables.py
```diff
@@ -364,11 +364,23 @@ def _validate_and_adjust_columns(self, table1, table2):
...
+
+                table1._schema[c] = col1.replace(precision=lowest.precision)
+                table2._schema[c] = col2.replace(precision=lowest.precision)
```
sirupsen, line 382: Can you explain why rounding isn't a problem for lower-precision floats and numerics, but is for DateTime? 👂🏻 

E.g. why do you have the problem where if you insert a datetime with `0.88888`, you run into a problem, but not with a `numeric(10, 1)`?

data_diff/database.py
```diff
@@ -212,23 +247,48 @@ def query(self, sql_ast: SqlOrStr, res_type: type):
...
+                return cls(precision=0)
+
+            elif issubclass(cls, Decimal):
+                return cls(precision=numeric_scale)
+
+            assert issubclass(cls, Float)
+            # assert numeric_scale is None
+            return cls(
+                precision=self._convert_db_precision_to_digits(
+                    numeric_precision if numeric_precision is not None else DEFAULT_NUMERIC_PRECISION
```
sirupsen, line 281: Why is 4-bytes float the safe default? Because even if it's 8-bytes, it is "at least" 4-bytes? 

Isn't this only a safe default if it's a float4 and `numeric_precision` isn't in the schema? This seems like code that should be in the driver it's relevant for, not the abstract class, no?

Should we maybe introduce `Double` in addition, so they can be typed out in `NUMERIC_TYPES` for each class?

tests/test_database_types.py
```diff
@@ -58,19 +65,31 @@
...
+            # "timestamp",
+            # "datetime(6)",
+        ],
         # https://dev.mysql.com/doc/refman/8.0/en/numeric-types.html
         "float": [
-            # "float",
-            # "double",
-            # "numeric",
+            "float",
+            "double",
```
sirupsen, line 78: aWhat about a test where we pass the precision bits to `float`?

data_diff/database.py
```diff
@@ -212,23 +247,48 @@ def query(self, sql_ast: SqlOrStr, res_type: type):
...
+                # We fill in the constant, so we need to ignore it for integers.
+                return cls(precision=0)
+
+            elif issubclass(cls, Decimal):
+                return cls(precision=numeric_scale)
+
+            assert issubclass(cls, Float)
+            # assert numeric_scale is None
+            return cls(
+                precision=self._convert_db_precision_to_digits(
```
sirupsen, line 280: For Postgres, `numeric_precision` in `INFORMATION_SCHEMA.columns` is 24, not 23 as I'd normally expect for a 4-byte float. Do you know why?

This causes a potential problem... For Postgres at least, the precision this function gives, and the documentation, aren't congruent for 4-byte, but is for 8-byte:

```python
math.floor(math.log(2**24, 10))
# => 7

>>> math.floor(math.log(2**53, 10))
15
```

CleanShot 2022-06-21 at 09 49 21@2x

https://www.postgresql.org/docs/current/datatype-numeric.html

data_diff/database.py
```diff
@@ -351,6 +421,9 @@ def normalize_value_by_type(self, value: str, coltype: ColType) -> str:
                 timestamp6 = f"to_char({value}::timestamp(6), 'YYYY-mm-dd HH24:MI:SS.US')"
                 return f"RPAD(LEFT({timestamp6}, {TIMESTAMP_PRECISION_POS+coltype.precision}), {TIMESTAMP_PRECISION_POS+6}, '0')"
 
+        elif isinstance(coltype, NumericType):
+            value = f"{value}::decimal(38, {coltype.precision})"
```
sirupsen, line 425: Why can you assume 38 digits in front of the `.`?

data_diff/database.py
```diff
@@ -362,6 +435,10 @@ class Presto(Database):
         "timestamp": Timestamp,
         # "datetime": Datetime,
     }
+    NUMERIC_TYPES = {
+        "integer": Integer,
+        "real": Float,
```
sirupsen, line 440: I think you're missing [double-precision](https://prestodb.io/docs/current/language/types.html#floating-point) here

data_diff/database.py
```diff
@@ -422,8 +502,24 @@ def _parse_type(self, type_repr: str, datetime_precision: int = None, numeric_pr
...
+                    rounds=False,
+                )
+
+        cls = self.NUMERIC_TYPES.get(type_repr)
+        if cls:
+            if issubclass(cls, Integer):
+                assert numeric_precision is not None
+                return cls(0)
+            elif issubclass(cls, Decimal):
+                return cls(6)
```
sirupsen, line 515: Why is this hardcoded for Decimal?

data_diff/database.py
```diff
@@ -433,6 +529,12 @@ class MySQL(ThreadedDatabase):
         "datetime": Datetime,
         "timestamp": Timestamp,
     }
+    NUMERIC_TYPES = {
+        "double": Float,
+        "float": Float,
+        "decimal": Decimal,
+        "int": Decimal,
```
sirupsen, line 536: Why is `int` decimal?

data_diff/database.py
```diff
@@ -532,14 +645,40 @@ def _parse_type(self, type_repr: str, datetime_precision: int = None, numeric_pr
...
-                return cls(precision=datetime_precision if datetime_precision is not None else DEFAULT_PRECISION,
-                    rounds=self.ROUNDS_ON_PREC_LOSS
+                return cls(
+                    precision=datetime_precision if datetime_precision is not None else DEFAULT_DATETIME_PRECISION,
+                    rounds=self.ROUNDS_ON_PREC_LOSS,
                 )
 
+        cls = {
+            "NUMBER": Decimal,
+            "FLOAT": Float,
```
sirupsen, line 655: What about `LONG`?

data_diff/database.py
```diff
@@ -671,6 +825,10 @@ class Snowflake(Database):
         "TIMESTAMP_LTZ": Timestamp,
         "TIMESTAMP_TZ": TimestampTZ,
     }
+    NUMERIC_TYPES = {
+        "NUMBER": Decimal,
+        "FLOAT": Float,
```
sirupsen, line 830: Snowflake has a ton of aliases, but they are all 4-bytes. I suspect that just works, because `numeric_precision` is just always 53 or whatever? 👍🏻 

CleanShot 2022-06-21 at 09 55 35@2x

tests/test_database_types.py
```diff
@@ -24,7 +31,7 @@
         "2022-05-01 15:10:03.003030",
         "2022-06-01 15:10:05.009900",
     ],
-    "float": [0.0, 0.1, 0.10, 10.0, 100.98],
+    "float": [0.0, 0.1, 0.00188, 0.99999, 0.091919, 0.10, 10.0, 100.98],
```
sirupsen, line 34: Some additional values that cause some more tests to fail:

```
     "float": [0.0, 0.1, 0.00188, 0.99999, 0.091919, 0.10, 10.0, 100.98,
         0.001201923076923077, 1/3, 1/5, 1/109, 1/109489, 1/1094893892389, 1/10948938923893289],
```

Failing tests:

* test_types_Snowflake_float_to_Postgres_real
* test_types_Redshift_float8_to_Postgres_real
* test_types_Postgres_real_to_Snowflake_float
* test_types_Postgres_real_to_Redshift_float8
* test_types_Postgres_real_to_Postgres_double_precision
* test_types_Postgres_real_to_MySQL_numeric_65_10_
* test_types_Postgres_double_precision_to_Postgres_real
* test_types_MySQL_numeric_65_10__to_Postgres_real

sirupsen (commented):
Really excited about this 😍 This and datetime should by far be the hardest, I hope!

tests/test_database_types.py
erezsh: I wish I knew. I just know that removing one of them causes some tests to fail.

data_diff/diff_tables.py
erezsh: It seems like all databases round their floats, instead of truncating them. Which makes sense as they all use the same FP algorithm.

But maybe we just didn't hit the right edge cases yet.

data_diff/database.py
erezsh: Yes.

> not the abstract class

See the refactor PR, it addresses this concern.

> Should we maybe introduce Double in addition

It isn't necessary for most DBs, because they provide numeric_scale.

For those that don't, yes, it can give us more accuracy. It just doesn't seem like high priority.

tests/test_database_types.py
erezsh: Maybe? I didn't make sure we're testing every variation.

data_diff/database.py
erezsh: No idea why..  Double still gives 53..

tests/test_database_types.py
sirupsen: 😭

data_diff/database.py
sirupsen: 👍🏻 I think it's worth a comment why this is a safe default, but yeah, I agree with you

data_diff/database.py
erezsh: Presto doesn't provide the numeric_scale. 6 is just a safe value.

data_diff/database.py
erezsh: These aliases are usually "normalized" to a single type by the dbs before stored in INFORMATION_SCHEMA.

tests/test_database_types.py
erezsh: Nice. I'll look into it!

erezsh: Working now.

data_diff/database.py
erezsh: Changed the way it's handled.

erezsh:
@sirupsen I think we're good. Anything else?

sirupsen (commented):
Let's goooo! 

Cut 0.1 when this is in master, and seed a changelog file? Can wait until you wake up tomorrow if you like, since the launch won't be until the morning eastern time I think.

PR #76, opened 2022-06-21
Better errors for missing imports

erezsh:


pyproject.toml
```diff
@@ -50,6 +50,7 @@ mysql = ["mysql-connector-python"]
 pgsql = ["psycopg2"]
```
sirupsen, line 50: Should we call this extra `postgres` instad? `pgsql` isn't even the name of the package 😅

pyproject.toml
```diff
@@ -50,6 +50,7 @@ mysql = ["mysql-connector-python"]
...
 snowflake = ["snowflake-connector-python"]
 presto = ["presto-python-client"]
+oracle = ["cx_Oracle"]
```
sirupsen, line 53: Ah good! What about BigQuery, Redshift, .. any others?

data_diff/database.py
```diff
@@ -22,6 +22,24 @@ def parse_table_name(t):
     return tuple(t.split("."))
 
 
+def import_helper(package: str = None, text=""):
```
sirupsen, line 25: Maybe we should add the extra name to the list of database tables in the README? Or at least in the `How to install` example, list all of them?

pyproject.toml
erezsh: I think it was to stay in the same format as mysql/mssql.

Anyway, I'm fine with either one.

sirupsen: Let's switch to `postgres` then

data_diff/database.py
sirupsen: I think it's the same issue: make it easier for people to figure out what drivers to install and when

sirupsen:
Let's get this in for 0.1 too

sirupsen (approved):

PR #88, opened 2022-06-21
readme: add performance graph

sirupsen:
The data comes from https://github.com/datafold/data-diff/pull/67 which I'll clean up significantly, then submit it for review. I will also need to add some instructions on how to use the benchmarking to generate these graphs

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,13 @@ there are few/no changes, but is able to output each differing row! By pushing
 the compute into the databases, it's _much_ faster than querying for and
 comparing every row.
 
+![Performance for 100M rows](https://user-images.githubusercontent.com/97400/174860361-35158d2b-0cad-4089-be66-8bf467058387.png)
+
+**†:** The implementation for downloading all rows that `data-diff` and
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -24,6 +24,12 @@ there are few/no changes, but is able to output each differing row! By pushing
 the compute into the databases, it's _much_ faster than querying for and
 comparing every row.
 
+![Performance for 100M rows](https://user-images.githubusercontent.com/97400/174860361-35158d2b-0cad-4089-be66-8bf467058387.png)
+
+**†:** Downloading all row implementation is not optimal. It's a single-process,
```
erezsh, line 29: This sentence needs a bit of work.

sirupsen: Can you tell me what you think needs work about it?

erezsh: Maybe just rephrase it:

The Python implementation for downloading rows isn't optimal in some of the drivers. It's a single process....

sirupsen: There's more to it than that, because of the GIL, etc. I did another pass and will merge

Issue #68, opened 2022-06-17
--stats gives weird info message


kning:
I get the message 
`Diff-Total: 3 changed rows out of None`

Looking at the code, it looks like None should be the count of records in table 1. My test case is a really small table (only 3 rows) so wondering if that's what's throwing things off.
![image](https://user-images.githubusercontent.com/1820651/174196842-00aa2b85-8d1f-4ba7-a54a-07f1f299f84b.png)

sirupsen:
This happens when all the rows were downloaded, not checksummed. Will fix

erezsh:
I believe @sirupsen fixed this.

PR #90, opened 2022-06-21
cli: output diff as jsonl, stats as json

sirupsen:
This makes it far easier to parse from a non-Python environment.

Would love this in for 0.1, as this is a pretty breaking change out of the gate

@erezsh

### DIFF
--- a/data_diff/__main__.py
+++ b/data_diff/__main__.py
@@ -51,7 +51,7 @@ COLOR_SCHEME = {
 @click.option("--max-age", default=None, help="Considers only rows younger than specified. See --min-age.")
 @click.option("-s", "--stats", is_flag=True, help="Print stats instead of a detailed diff")
 @click.option("-d", "--debug", is_flag=True, help="Print debug info")
-@click.option("--json", 'json_output', is_flag=True, help="Print JSON output for --stats")
+@click.option("--json", 'json_output', is_flag=True, help="Print JSONL output for machine readability")
 @click.option("-v", "--verbose", is_flag=True, help="Print extra info")
 @click.option("-i", "--interactive", is_flag=True, help="Confirm queries, implies --debug")
 @click.option("--keep-column-case", is_flag=True, help="Don't use the schema to fix the case of given column names.")
@@ -160,16 +160,22 @@ def main(
                 "different_-": minus,
                 "total": max_table_count,
             }
-            print(json.dumps(json_output, indent=2))
+            print(json.dumps(json_output))
         else:
             print(f"Diff-Total: {len(diff)} changed rows out of {max_table_count}")
             print(f"Diff-Percent: {percent:.14f}%")
             print(f"Diff-Split: +{plus}  -{minus}")
     else:
-        for op, key in diff_iter:
+        for op, columns in diff_iter:
             color = COLOR_SCHEME[op]
-            jsonl = json.dumps([op, list(key)])
-            rich.print(f"[{color}]{jsonl}[/{color}]")
+
+            if json_output:
+                jsonl = json.dumps([op, list(columns)])
+                rich.print(f"[{color}]{jsonl}[/{color}]")
+            else:
+                text = f"{op} {', '.join(columns)}"
+                rich.print(f"[{color}]{text}[/{color}]")
+
             sys.stdout.flush()
 
     end = time.time()
--- a/data_diff/diff_tables.py
+++ b/data_diff/diff_tables.py
@@ -382,11 +382,12 @@ class TableDiffer:
             rows1, rows2 = self._threaded_call("get_values", [table1, table2])
             diff = list(diff_sets(rows1, rows2))
 
-            # This happens when the initial bisection threshold is larger than
-            # the table itself.
-            if level == 0 and not self.stats.get("table1_count", False):
-                self.stats["table1_count"] = self.stats.get("table1_count", 0) + len(rows1)
-                self.stats["table2_count"] = self.stats.get("table2_count", 0) + len(rows2)
+            # Initial bisection_threshold larger than count. Normally we always
+            # checksum and count segments, even if we get the values. At the
+            # first level, however, that won't be true.
+            if level == 0:
+                self.stats["table1_count"] = len(rows1)
+                self.stats["table2_count"] = len(rows2)
 
             logger.info(". " * level + f"Diff found {len(diff)} different rows.")
             self.stats["rows_downloaded"] = self.stats.get("rows_downloaded", 0) + max(len(rows1), len(rows2))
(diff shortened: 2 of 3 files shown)


data_diff/diff_tables.py
```diff
@@ -381,6 +381,12 @@ def _bisect_and_diff_tables(self, table1, table2, level=0, max_rows=None):
         if max_rows < self.bisection_threshold:
             rows1, rows2 = self._threaded_call("get_values", [table1, table2])
             diff = list(diff_sets(rows1, rows2))
+
+            # This happens when the initial bisection threshold is larger than
+            # the table itself.
+            if level == 0 and not self.stats.get("table_count", False):
+                self.stats["table_count"] = self.stats.get("table_count", 0) + max(len(rows1), len(rows2))
```
sirupsen, line 388: This fixes https://github.com/datafold/data-diff/issues/68

When the count is lower than the bisection threshold on the initial segment, the table count is never set.

data_diff/__main__.py
```diff
@@ -146,15 +147,24 @@ def main(
...
+        count = differ.stats["table_count"]
+        diff = {
+            "different_rows": len(diff),
+            "different_percent": percent,
+            "different_+": plus,
+            "different_-": minus,
+            "total": count,
+        }
+
+        print(json.dumps(diff, indent=2))
```
sirupsen, line 162: I know this is a bit less pretty, but I think this is most suitable as a tool that we want people to plug into all kinds of ecosystems.

data_diff/diff_tables.py
```diff
@@ -420,7 +426,7 @@ def _diff_tables(self, table1, table2, level=0, segment_index=None, segment_coun
             return
 
         if level == 1:
-            self.stats["table1_count"] = self.stats.get("table1_count", 0) + count1
+            self.stats["table_count"] = self.stats.get("table_count", 0) + max(count1, count2)
```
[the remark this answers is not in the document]
erezsh: Maybe we should count both tables?

data_diff/__main__.py
erezsh: The default output should be easy to read for humans

We can have a `--json` switch for JSON output.

data_diff/diff_tables.py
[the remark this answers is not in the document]
sirupsen: You mean independently? And then `max` them in the CLI? Sure, I don't really care since it's the only user. I guess since it's exposed via the API we can do that

data_diff/__main__.py
```diff
@@ -144,17 +147,29 @@ def main(
...
+                "different_rows": len(diff),
+                "different_percent": percent,
+                "different_+": plus,
+                "different_-": minus,
+                "total": max_table_count,
+            }
+            print(json.dumps(json_output, indent=2))
+        else:
+            print(f"Diff-Total: {len(diff)} changed rows out of {max_table_count}")
+            print(f"Diff-Percent: {percent:.14f}%")
```
sirupsen, line 166: BTW I stopped rounding aggressively here because e.g. when I have 1 different row for 100m entries, it just shows 0 with that previous rounding

sirupsen:
Let me know if this is good to go now @erezsh

data_diff/__main__.py
```diff
@@ -146,15 +147,24 @@ def main(
...
+            "total": count,
+        }
+
+        print(json.dumps(diff, indent=2))
     else:
         for op, key in diff_iter:
             color = COLOR_SCHEME[op]
-            rich.print(f"[{color}]{op} {key!r}[/{color}]")
+            jsonl = json.dumps([op, list(key)])
+            rich.print(f"[{color}]{jsonl}[/{color}]")
```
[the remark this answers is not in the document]
sirupsen: Let's just default to JSON then. I don't see why we'd need two formats here.

data_diff/diff_tables.py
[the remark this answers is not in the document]
sirupsen: Done

sirupsen:
@erezsh can you review my replies?

data_diff/__main__.py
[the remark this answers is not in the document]
erezsh: image

VS

image

You can try to argue they are "just as readable" but there is no way you will convince me.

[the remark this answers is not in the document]
erezsh: (Produced with `text = f"{op} {', '.join(columns)}"`)

data_diff/diff_tables.py
```diff
@@ -381,6 +381,13 @@ def _bisect_and_diff_tables(self, table1, table2, level=0, max_rows=None):
         if max_rows < self.bisection_threshold:
             rows1, rows2 = self._threaded_call("get_values", [table1, table2])
             diff = list(diff_sets(rows1, rows2))
+
+            # This happens when the initial bisection threshold is larger than
+            # the table itself.
+            if level == 0 and not self.stats.get("table1_count", False):
```
erezsh, line 387: Good catch. But I think we always want to override these attribute. Why would we care what previous values they held?

sirupsen: (talked about on slack, just for the record) Will double count if it's not level 0, but we can remove the not

data_diff/__main__.py
```diff
@@ -50,6 +51,7 @@
 @click.option("--max-age", default=None, help="Considers only rows younger than specified. See --min-age.")
 @click.option("-s", "--stats", is_flag=True, help="Print stats instead of a detailed diff")
 @click.option("-d", "--debug", is_flag=True, help="Print debug info")
+@click.option("--json", 'json_output', is_flag=True, help="Print JSON output for machine readability")
```
erezsh, line 54: JSONL

data_diff/__main__.py
```diff
@@ -144,17 +147,34 @@ def main(
...
     else:
         for op, key in diff_iter:
             color = COLOR_SCHEME[op]
-            rich.print(f"[{color}]{op} {key!r}[/{color}]")
+
+            if json_output:
+                jsonl = json.dumps([op, list(key)])
+                rich.print(f"[{color}]{jsonl}[/{color}]")
+            else:
+                rich.print(f"[{color}]{op} {key!r}[/{color}]")
```
erezsh, line 176: Can you change the "human output" to this?

```python
text = f"{op} {', '.join(columns)}")
```

data_diff/diff_tables.py
```diff
@@ -381,6 +381,14 @@ def _bisect_and_diff_tables(self, table1, table2, level=0, max_rows=None):
         if max_rows < self.bisection_threshold:
             rows1, rows2 = self._threaded_call("get_values", [table1, table2])
             diff = list(diff_sets(rows1, rows2))
+
+            # Initial bisection_threshold larger than count. Normally we always
+            # checksum and count segments, even if we get the values. At the
+            # first level, however, that won't be true.
+            if level == 0:
+                self.stats["table1_count"] = self.stats.get("table1_count", 0) + len(rows1)
+                self.stats["table2_count"] = self.stats.get("table2_count", 0) + len(rows2)
```
erezsh, line 390: Still this line... just 
```
self.stats["table1_count"] = len(rows1)
```

No?

PR #91, opened 2022-06-21
improve support for Presto and Snowflake 

sfenner-eb:
Snowflake:
Allow auth by PKCS8 key for Snowflake (password auth disabled)
User private key specification
Mapping bigint to decimal

Presto:
Allow verify http session with certificate for Presto
Allow using basic authentication for Presto
Presto Port specification
Presto HTTP scheme specification
Presto certificate specification
Mapping bigint to Integer

### DIFF
--- a/data_diff/database.py
+++ b/data_diff/database.py
@@ -9,6 +9,8 @@ from typing import Tuple, Optional, List
 from concurrent.futures import ThreadPoolExecutor
 import threading
 from typing import Dict
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives import serialization
 
 import dsnparse
 import sys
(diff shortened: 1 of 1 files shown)


sfenner-eb:
example using key for snowflake and multiple options for presto (https, basic authenticator, pass cert to verify)

` poetry run python3 -m data_diff 'snowflake://<user>@<account>/db?warehouse=<warehouse>&role=<role>&schema=<schema>&key=<path_to_key.p8>'  table 'presto://<user>:<password>@<host>:<port>/<catalog>/<schema>?auth=basic&http_scheme=https&source=<resource_group>&cert=<path_to_.cer>' table `

sfenner-eb:
poetry run python3 -m data_diff 'snowflake://yyy@vvv.COM@vvv.us-east-1/dev_source_yyy?warehouse=dev_yyy_small&role=dev-yyy-zzz&schema=riskyy&key=/Users/yyyy/.ssh/jjj_rsa_key.p8' RISKY.REVIEW_EMAILS 'presto://yyy:pw@presto.env.team.vv:8443/hive/risky?auth=basic&http_scheme=https&source=odbc&cert=/Users/yyyy/vvca.cer' review_emails  --key-column ID --columns 'CREATED' --columns 'SUBJECT' --columns 'REASON' --columns 'TYPE' --columns 'STATUS'--bisection-factor=50 --bisection-threshold 25000 -v -d

actual command (...well, you know).


Using data-diff  to automatically verify import of data from hive/s3 (Presto) into Snowflake.

data_diff/database.py
```diff
@@ -852,6 +862,7 @@ class Snowflake(Database):
     NUMERIC_TYPES = {
         "NUMBER": Decimal,
         "FLOAT": Float,
+        "BIGINT": Decimal,
```
erezsh, line 865: I believe this should be `Integer`.

Did you add this because you got an error in the tests?

data_diff/database.py
```diff
@@ -9,6 +9,8 @@
 from concurrent.futures import ThreadPoolExecutor
 import threading
 from typing import Dict
+from cryptography.hazmat.backends import default_backend
```
erezsh, line 12: There is no guarantee that this package exists.

1) Included in pyproject for when running `pip install data-diff[snowflake]`

2) Imported only when snowflake is loaded. (See `import_snowflake()`)

erezsh (changes requested):
Thanks for the PR! Sorry it took me a while to review.

It looks okay, although I currently have no way to test if the code works or not. I'll see if we can adapt the test env to include this kind of auth.

Also, we refactored `database.py` into `databases/*`, where each database gets its own filename. Sorry for the trouble, but please re-implement this PR over the latest `master`. It was a one time "breaking" change.

One last thing, please run `black -l 120` before pushing, as it's the formatting standard for this project.

sirupsen:
@sfenner-eb in addition, can you provide some details on how we can test this? For Presto, you should be able to change the `docker-compose` setup to test with the new options.

For Snowflake, we're less familiar; can you tell us how we can test it?

PR #97, opened 2022-06-22
readme: update screenshot with BigQuery, remove redundant pair

sirupsen:
@erezsh this removes `MySQL → Postgres: int` (equivalent to `Postgres → MySQL: int`) and adds `Postgres → BigQuery`

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ there are few/no changes, but is able to output each differing row! By pushing
 the compute into the databases, it's _much_ faster than querying for and
 comparing every row.
 
-![Performance for 100M rows](https://user-images.githubusercontent.com/97400/174860361-35158d2b-0cad-4089-be66-8bf467058387.png)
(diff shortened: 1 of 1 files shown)


erezsh:
Nice. Maybe we should sort them by something? Either by int/timestamp or by cloud/not-cloud?

sirupsen:
Alright, sorted it by whether they're in the cloud, and timestamps last. I think this makes sense 👍🏻 

newplot (8)

README.md
```diff
@@ -185,7 +185,7 @@ Options:
   - `-v` or `--verbose` - Print extra info
   - `-i` or `--interactive` - Confirm queries, implies `--debug`
   - `--json` - Print JSONL output for machine readability
-  - `--min-age` - Considers only rows older than specified.
+  - `--min-age` - Considers only rows older than specified. Useful for specifying replication lag.
```
sirupsen, line 188: You cool with this?

erezsh: Yep.

erezsh (approved):

PR #106, opened 2022-06-24
tests: fix presto, add faker classes

sirupsen:
First set of changes to allow the test suites to be used for benchmarking:

1. _Actually_ use Presto for inserting/etc., before it was just using the backing Postgres table I think, because it wasn't working properly for me (and fix some Presto related bugs)
2. Add `*Faker` classes that are capable of producing an infinite amount of rows for benchmarking
3. Add pagination for the import into the destination table, required when we do millions of rows
4. Fix Presto rounding (I think this is supporting of (1) that it wasn't tested the way we thought?)
5. Add a few more things to make the test and table names nicer

Just trying to split this up so it's easier to review

### DIFF
--- a/tests/common.py
+++ b/tests/common.py
@@ -1,9 +1,8 @@
 import hashlib
+import os
 
 from data_diff import databases as db
 import logging
-
-logging.basicConfig(level=logging.INFO)
 
 TEST_MYSQL_CONN_STRING: str = "mysql://mysql:Password1@localhost/mysql"
 TEST_POSTGRESQL_CONN_STRING: str = None
@@ -13,6 +12,16 @@ TEST_REDSHIFT_CONN_STRING: str = None
 TEST_ORACLE_CONN_STRING: str = None
 TEST_PRESTO_CONN_STRING: str = None
 
+DEFAULT_N_SAMPLES = 50
+N_SAMPLES = int(os.environ.get("N_SAMPLES", DEFAULT_N_SAMPLES))
+
+level = logging.ERROR
+if os.environ.get("LOG_LEVEL", False):
+    level = getattr(logging, os.environ["LOG_LEVEL"].upper())
+
+logging.basicConfig(level=level)
+logging.getLogger("diff_tables").setLevel(level)
+logging.getLogger("database").setLevel(level)
 
 try:
     from .local_settings import *
--- a/tests/test_database_types.py
+++ b/tests/test_database_types.py
@@ -1,34 +1,120 @@
 from contextlib import suppress
 import unittest
 import time
-import logging
+import re
+import math
+import datetime
 from decimal import Decimal
-
 from parameterized import parameterized
 
 from data_diff import databases as db
 from data_diff.diff_tables import TableDiffer, TableSegment
-from .common import CONN_STRINGS
+from .common import CONN_STRINGS, N_SAMPLES
 
 
-logging.getLogger("diff_tables").setLevel(logging.ERROR)
-logging.getLogger("database").setLevel(logging.WARN)
-
-CONNS = {k: db.connect_to_uri(v) for k, v in CONN_STRINGS.items()}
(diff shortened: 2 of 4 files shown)


tests/test_database_types.py
```diff
@@ -203,18 +313,42 @@
...
                         )
 
+def sanitize(name):
+    name = name.lower()
+    name = re.sub(r"[\(\)]", "", name)  #  timestamp(9) -> timestamp9
+    # Try to shorten long fields, due to length limitations in some DBs
+    name = name.replace(r"without time zone", "n_tz")
+    name = name.replace(r"with time zone", "y_tz")
+    name = name.replace(r"with local time zone", "y_tz")
+    name = name.replace(r"timestamp", "ts")
```
sirupsen, line 323: I prefer shortening and never truncating names. That could lead to some nasty, nasty bugs down the line.

tests/common.py
```diff
@@ -13,6 +12,16 @@
 TEST_ORACLE_CONN_STRING: str = None
 TEST_PRESTO_CONN_STRING: str = None
 
+DEFAULT_N_SAMPLES = 50
+N_SAMPLES = int(os.environ.get('N_SAMPLES', DEFAULT_N_SAMPLES))
+
+level = logging.ERROR
+if os.environ.get('DEBUG', False):
+    level = logging.DEBUG
```
erezsh, line 20: It would be more flexible to get the `LOG_LEVEL` and use getattr with it.

tests/test_database_types.py
```diff
@@ -2,33 +2,117 @@
 import unittest
 import time
 import logging
+import re
+import math
+import datetime
 from decimal import Decimal
+from prestodb.exceptions import PrestoUserError
```
erezsh, line 9: I think we should avoid this kind of imports.

Instead you can use `import_presto()` from `databases/presto.py` where it's needed.

If that's too clumsy, we can store "prestodb" in the Presto insntance, and use it as `db.prestodb.exceptions.PrestoUserError`.

tests/test_database_types.py
```diff
@@ -2,33 +2,117 @@
...
+
+    def __iter__(self):
+        self.last_id = 0
+        self.values = []
+        self.value_index = 0
+        return self
+
+    def __next__(self) -> str:
+        if self.value_index == len(self.values):  #  end of current batch
+            query = f"SELECT id, col FROM {self.table} WHERE id > {self.last_id} ORDER BY id ASC LIMIT {self.RECORDS_PER_BATCH}"
```
erezsh, line 39: Oracle doesn't support LIMIT. Instead you have to do `FETCH NEXT n ROWS ONLY`

tests/test_database_types.py
erezsh: That's a nice idea!

tests/test_database_types.py
```diff
@@ -2,33 +2,117 @@
...
+class PaginatedTable:
+    # We can't query all the rows at once for large tables. It'll occupy too
+    # much memory.
+    RECORDS_PER_BATCH = 1000000
+
+    def __init__(self, table, conn):
+        self.table = table
+        self.conn = conn
+
+    def __iter__(self):
```
erezsh, line 31: I think this kind of mutable iterator where you always return `self` in `__iter__` isn't thread safe. And these tests are much better when they can be run in parallel.

To make it thread-safe, `__iter__` should return a new `PaginatedTableIter` which keeps all the mutable attributes contained.

If you want to test it, use a package called `unittest-parallel`.

erezsh (changes requested):

erezsh:
Nice one!

tests/test_database_types.py
sirupsen: See the line below!

erezsh: Good enough. I'll refactor that later into `sql.py`.

erezsh (approved):

Issue #115, opened 2022-06-27
Specify schema name for Oracle


mpcarter:
Trying to test a comparison to an Oracle database, but I cannot figure out how to inform data-diff which schema the table is in. It only seems to support specifying a table name and thus only tables belonging to the user I am logged in as. 

Our DBAs do not let users log in using the owner of our production schemas (for good reason).

erezsh:
Thanks for the information! That should be pretty simple to do.

Do I understand correctly that the information in USER_TAB_COLUMNS is essentially the same as ALL_TAB_COLUMNS  where `SCHEMA_NAME = <user-name>`?

mpcarter:
Yes that is correct.

erezsh:
@mpcarter I created a PR that I believe corrects this issue (#117 ). If you can give it a try and let me know if it works for you, that would help fast-tracking it into the master branch.

For convenience, you can install it with `pip install git+https://github.com/datafold/data-diff.git@refs/pull/117/head`.

mpcarter:
@erezsh Not sure if you prefer comments here or on the PR. I tried it out and it worked in as much as it was able to query the correct table. However, another error occurred. I ran it debug flag and included output below.

```
(datadiffpr117) datadiff $ data-diff $ORA MYSCHEMA.MY_TABLE $SF MY_TABLE -k ID -d
[08:18:52] DEBUG - Injecting ssl_wrap_socket_with_ocsp
[08:18:52] DEBUG - cache directory: C:\Users\mcarter\AppData\Local\Snowflake\Caches
[08:18:52] DEBUG - Failed to import pyarrow. Cannot use pandas fetch API
Initiating login request with your identity provider. A browser window should have opened for you to complete the login. If you can't see it, check existing browser windows, or your OS settings. Press CTRL+C to abort and try again...
[08:18:53] DEBUG - Starting new HTTP connection (1): ocsp.snowflakecomputing.com:80
[08:18:53] DEBUG - http://ocsp.snowflakecomputing.com:80 "GET /ocsp_response_cache.json HTTP/1.1" 200 181522
[08:18:56] DEBUG - Running SQL (Oracle): SELECT column_name, data_type, 6 as datetime_precision, data_precision as numeric_precision, data_scale as numeric_scale FROM ALL_TAB_COLUMNS WHERE table_name = 'MY_TABLE' AND owner = 'MYSCHEMA'
[...]

PR #117, opened 2022-06-28
Oracle: Added support for schemas (Issue #115)

erezsh:


sirupsen (approved):
LGTM

Another problem he seemed to have was `concat` not supporting just 1 argument in Oracle, can you add this patch too to `sql.py`?

```python
    def compile(self, c: Compiler):
        compiled_exprs = ", ".join(map(c.compile, self.exprs))
        expr = compiled_exprs
        if len(self.exprs) > 1:
            expr = f"concat({compiled_exprs})"
        md5 = c.database.md5_to_int(expr)
        return f"sum({md5})"
```

PR #119, opened 2022-06-28
Support for UUID key column

erezsh:
Currently only working for MySQL, but I'm posting it for comments about my approach and choices.

Next steps would be to -

- Add more tests. Maybe into test_database_types

- Support more databases and column types.

### DIFF
--- a/data_diff/databases/base.py
+++ b/data_diff/databases/base.py
@@ -1,13 +1,15 @@
+from uuid import UUID
 import math
 import sys
 import logging
-from typing import Dict, Tuple, Optional, Sequence, Type
+from typing import Dict, Tuple, Optional, Sequence, Type, List
 from functools import lru_cache, wraps
 from concurrent.futures import ThreadPoolExecutor
 import threading
 from abc import abstractmethod
 
 from .database_types import (
+    ColType_UUID,
     AbstractDatabase,
     ColType,
     Integer,
@@ -16,14 +18,23 @@ from .database_types import (
     PrecisionType,
     TemporalType,
     UnknownColType,
+    Text,
 )
-from data_diff.sql import DbPath, SqlOrStr, Compiler, Explain, Select
+from data_diff.sql import DbPath, SqlOrStr, Compiler, Explain, Select, TableName
 
 logger = logging.getLogger("database")
 
 
 def parse_table_name(t):
     return tuple(t.split("."))
+
+
+def is_uuid(u):
+    try:
+        UUID(u)
+    except ValueError:
+        return False
+    return True
 
 
 def import_helper(package: str = None, text=""):
@@ -102,7 +113,7 @@ class Database(AbstractDatabase):
             assert len(res) == 1, (sql_code, res)
             return res[0]
         elif getattr(res_type, "__origin__", None) is list and len(res_type.__args__) == 1:
-            if res_type.__args__ == (int,):
+            if res_type.__args__ == (int,) or res_type.__args__ == (str,):
                 return [_one(row) for row in res]
             elif res_type.__args__ == (Tuple,):
                 return [tuple(row) for row in res]
@@ -123,6 +134,7 @@ class Database(AbstractDatabase):
 
     def _parse_type(
         self,
+        table_path: DbPath,
         col_name: str,
         type_repr: str,
         datetime_precision: int = None,
@@ -147,7 +159,7 @@ class Database(AbstractDatabase):
         elif issubclass(cls, Decimal):
             if numeric_scale is None:
                 raise ValueError(
-                    f"{self.name}: Unexpected numeric_scale is NULL, for column {col_name} of type {type_repr}."
+                    f"{self.name}: Unexpected numeric_scale is NULL, for column {'.'.join(table_path)}.{col_name} of type {type_repr}."
                 )
             return cls(precision=numeric_scale)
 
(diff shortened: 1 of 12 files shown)


data_diff/sql.py
```diff
@@ -93,6 +96,9 @@ def compile(self, parent_c: Compiler):
         if self.order_by:
             select += " ORDER BY " + ", ".join(map(c.compile, self.order_by))
 
+        if self.limit is not None:
+            select += " " + c.database.offset_limit(0, self.limit)
```
sirupsen, line 100: Great you added support for this! Can you change the `test_database_types.py` to use it?

data_diff/databases/database_types.py
```diff
@@ -1,11 +1,32 @@
+from uuid import UUID
 from abc import ABC, abstractmethod
-from typing import Sequence, Optional, Tuple, Union, Dict
+from typing import Sequence, Optional, Tuple, Union, Dict, Any
 from datetime import datetime
 
 from runtype import dataclass
 
+
+class ArithUUID(UUID):
```
sirupsen, line 9: This is sweet

tests/test_diff_tables.py
```diff
@@ -239,6 +240,43 @@ def test_diff_sorted_by_key(self):
...
+        for query in queries:
+            self.connection.query(query, None)
+
+        self.a = TableSegment(self.connection, ("a",), "id", "comment")
+        self.b = TableSegment(self.connection, ("b",), "id", "comment")
+
+    def test_string_keys(self):
+        differ = TableDiffer()
+        diff = list(differ.diff_tables(self.a, self.b))
+        self.assertEqual(diff, [("-", (str(self.new_uuid), "This one is different"))])
```
sirupsen, line 277: Some more tests here would be great, e.g.:

* Some rows are UUIDS, others are not
* Invalid or weird UUIDs (is that possible?)
* Maybe also some where there's a few more differences and no differences?

sirupsen (commented):
Nice! This is a really elegant implementation.

Would love to see those type tests you mentioned in the opening description to make sure it behaves. Probably we can have a `UUIDFaker` or something. We might need both a "string" and "uuid" type category to test this there?

data_diff/databases/database_types.py
erezsh: Yeah, the magic of Python. It wouldn't be possible in most languages.

tests/test_diff_tables.py
erezsh: Right now, if I see anything that's not UUID, I just throw an error. But sure, I can test that.

data_diff/databases/oracle.py
```diff
@@ -13,6 +13,10 @@ def import_oracle():
 
 
 class Oracle(ThreadedDatabase):
+    TYPE_CLASSES: Dict[str, type] = {
+        "NUMBER": Decimal,
+        "FLOAT": Float,
```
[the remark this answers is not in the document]
erezsh: I added them to the tests, but in the schema they always appear as FLOAT.

data_diff/databases/base.py
```diff
@@ -119,36 +143,49 @@ def _parse_type(
...
             # assert numeric_scale is None
             return cls(
                 precision=self._convert_db_precision_to_digits(
                     numeric_precision if numeric_precision is not None else DEFAULT_NUMERIC_PRECISION
                 )
             )
 
-        return UnknownColType(type_repr)
+        elif issubclass(cls, Text):
+            samples = self.query(Select([col_name], TableName(table_path), limit=16), List[str])
```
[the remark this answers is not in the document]
sirupsen: I was just thinking you could do a `SELECT *` and then cache the samples 👀

data_diff/databases/oracle.py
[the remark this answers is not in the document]
sirupsen: Ooooh!

erezsh:
@sirupsen This PR is ready. I added a few more changes, but nothing too surprising.

Still, thought I'd give you a chance to review it.

tests/test_database_types.py
```diff
@@ -394,7 +442,8 @@ def _drop_table_if_exists(conn, table):
             conn.query(f"DROP TABLE {table}", None)
         else:
             conn.query(f"DROP TABLE IF EXISTS {table}", None)
-            conn.query("COMMIT", None)
+            if not isinstance(conn, db.BigQuery):
+                conn.query("COMMIT", None)
```
sirupsen, line 446: Ah yeah, because of that weird problem where we don't run both 😱 

Can you add an issue to fix that?

data_diff/diff_tables.py
```diff
@@ -105,6 +97,7 @@ def with_schema(self) -> "TableSegment":
             return self
 
         schema = self.database.query_table_schema(self.table_path, self._relevant_columns)
+        logger.debug(f"[{self.database.name}] Schema = {schema}")
```
sirupsen, line 100: 🤘🏻

data_diff/diff_tables.py
```diff
@@ -346,12 +348,16 @@ def _validate_and_adjust_columns(self, table1, table2):
                 table1._schema[c] = col1.replace(precision=lowest.precision)
                 table2._schema[c] = col2.replace(precision=lowest.precision)
 
+            elif isinstance(col1, StringType):
+                if not isinstance(col2, StringType):
+                    raise TypeError(f"Incompatible types for column '{c}':  {col1} <-> {col2}")
```
sirupsen, line 353: Will repeat my pet-peeve of always including database names, because I've seen this error and scratched my head so many times 😅 I know it's a bit less urgent here, but a good rule for us to follow I think. Maybe something like `Postgres/varchar(100)` would probably be fine.

data_diff/databases/database_types.py
```diff
@@ -50,11 +55,36 @@ class Float(FractionalType):
...
+    "Interface for ColType, for using a column as a key in data-diff"
+    python_type: type
+
+
+class ColType_UUID(StringType, IKey):
+    python_type = ArithUUID
+
+
+@dataclass
+class Text(StringType):
```
sirupsen, line 79: Let's add an issue to support this type 👍🏻

sirupsen (approved):
Looking good Erez!! Let's get them to check this out in the issue!

Only some small comments that I'll leave at your discretion, won't need to look again

data_diff/diff_tables.py
erezsh: Yeah, I agree. In general I should do another pass on error messages.

Issue #116, opened 2022-06-27
readme should clarify that DB_URI arguments need to be surrounded in quotes when connecting to Snowflake


leoebfolsom:
I figured I'd share a 🤦  experience I just had during my setup with Snowflake. I was having trouble getting data-diff to accept my `DB1_URI` and `DB2_URI` arguments. I eventually realized they had to be in quotes. One would not have known that from [the README](https://github.com/datafold/data-diff#example-command-and-output) which offers a postgres example that doesn't put the arguments in quotes (I assume quotes are not needed when using postgres?).

Here is what I encountered (I intentionally omitted arguments for brevity and RCA while debugging).

Without quotes:

```
% data-diff snowflake://me:mypw@myaccount/mydb/myschema?warehouse=mywh&role=myrole 
[1] 12578
zsh: no matches found: snowflake://me:mypw@myaccount/mydb/myschema?warehouse=mywh
% 
[1]  + exit 1     data-diff snowflake://me:mypw@myaccount/mydb/myschema?warehouse=mywh
```

(huh? did I enter my snowflake info wrong? scratches head)

With quotes:
```
% data-diff "snowflake://me:mypw@myaccount/mydb/myschema?warehouse=mywh&role=myrole"
Usage: data-diff [OPTIONS] DB1_URI TABLE1_NAME DB2_URI
                 TABLE2_NAME
Try 'data-diff --help' for help.

Error: Missing argument 'TABLE1_NAME'.
```

(nice! the error message helpfully tells me that I should enter the other arguments) :cool:

I think it would help a lot if right under the postgres example, there was a similar example for Snowflake, and imo for each tool that is up and running, especially since the syntax evidently varies.

I don't think every example needs to be as verbose as the postgres example, which includes output, but capturing any variation in syntax seems crucial. I don't think it's reasonable to expect users to know that the syntax for arguments differs.

The Snowflake example would look like this to make it equivalent to the postres example:

```
$ data-diff \
    "snowflake://<user>:<password>@<myaccount>/<database>/<schema>?warehouse=<warehouse>&role=<role>" RATING \
    "snowflake://<user>:<password>@<myaccount>/<database>/<schema>?warehouse=<warehouse>&role=<role>" RATING_DEL1 \
    --bisection-threshold 100000 \ # for readability, try default first
    --bisection-factor 6 \ # for readability, try default first
    --update-column timestamp \
    --verbose
```

A couple of other notes:

[...]

PR #120, opened 2022-06-28
lf/update readme snowflake

leoebfolsom:
Small edits to the README [as described/discussed here](https://github.com/datafold/data-diff/issues/116).

- quotes around Snowflake URI
- a note about capitalizing Snowflake table and schema names, and a general advisory about case sensitivity
- 🥕 around user-inputted arguments in the database table

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,13 +112,13 @@ $ data-diff \
 
 | Database      | Connection string                                                                       | Status |
 |---------------|-----------------------------------------------------------------------------------------|--------|
-| PostgreSQL    | `postgresql://user:password@hostname:5432/database`                                     |  💚    |
-| MySQL         | `mysql://user:password@hostname:5432/database`                                          |  💚    |
-| Snowflake     | `snowflake://user:password@account/database/SCHEMA?warehouse=WAREHOUSE&role=role`       |  💚    |
-| Oracle        | `oracle://username:password@hostname/database`                                          |  💛    |
-| BigQuery      | `bigquery://project/dataset`                                                            |  💛    |
-| Redshift      | `redshift://username:password@hostname:5439/database`                                   |  💛    |
-| Presto        | `presto://username:password@hostname:8080/database`                                     |  💛    |
+| PostgreSQL    | `postgresql://<user>:<password>@<hostname>:<5432>/<database>`                           |  💚    |
+| MySQL         | `mysql://<user>:<password>@<hostname>:<5432>/<database>`                                |  💚    |
+| Snowflake     | `"snowflake://<user>:<password>@<account>/<database>/<SCHEMA>?warehouse=<WAREHOUSE>&role=<role>"`|  💚    |
+| Oracle        | `oracle://<username>:<password>@<hostname>/database`                                          |  💛    |
+| BigQuery      | `bigquery://<project>/<dataset>`                                                            |  💛    |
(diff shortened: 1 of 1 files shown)


sirupsen (approved):
Just the one comment, otherwise LGTM. Since get params aren't used in the others, I agree it's not necessary.

@erezsh should sign off too

README.md
```diff
@@ -171,7 +171,10 @@ Users can also install several drivers at once:
 Usage: `data-diff DB1_URI TABLE1_NAME DB2_URI TABLE2_NAME [OPTIONS]`
 
 See the [example command](#example-command-and-output) and the [sample
-connection strings](#supported-databases).
+connection strings](#supported-databases). 
+
+Note that for some databases, the arguments that you enter in the command line 
+may be case-sensitive. This is the case for the Snowflake schema and table names.
```
erezsh, line 177: As far as I know, only the schema name is case-sensitive.

We should handle case insensitive table names by default.

erezsh (changes requested):
Overall, I'm okay with this PR.

See one comment that needs to be addressed.

README.md
sirupsen: @erezsh I thought we only handled case insensitivity on columns? Don't think we query the DB for the tables, do we?

erezsh (approved):

README.md
leoebfolsom: Hi @erezsh, [this](https://github.com/datafold/data-diff/blob/43c1c24bdac06e2753067d6671e23cb595596306/data_diff/databases/base.py#L157) is where I ran into the issue (you probably knew that).

In Snowflake, the following queries return 0 rows due to incorrect casing:
```
SELECT column_name, data_type, datetime_precision, numeric_precision, numeric_scale FROM information_schema.columns 
WHERE table_name = 'deal_facts' AND table_schema = 'DBT_LFOLSOM';
```
```
SELECT column_name, data_type, datetime_precision, numeric_precision, numeric_scale FROM information_schema.columns 
WHERE table_name = 'DEAL_FACTS' AND table_schema = 'dbt_lfolsom';
```
```
SELECT column_name, data_type, datetime_precision, numeric_precision, numeric_scale FROM information_schema.columns 
WHERE table_name = 'deal_facts' AND table_schema = 'dbt_lfolsom';
```
Only this version returns rows:
```
SELECT column_name, data_type, datetime_precision, numeric_precision, numeric_scale FROM information_schema.columns 
WHERE table_name = 'DEAL_FACTS' AND table_schema = 'DBT_LFOLSOM';
```

_Both_ `DEAL_FACTS` AND `DBT_LFOLSOM` have to be all-caps to match how they are named in the db. 

If I enter the command line arguments as lowercase, the query that is sent to Snowflake respects whatever case I entered, leading the query to fail.

So I believe both table and schema names are case-sensitive in Snowflake. 

I also think it is possible to ask Snowflake to store table or schema names with a particular case (lower, upper, or other), so it is unpredictable what should be sent to Snowflake. 

Personally, I like the idea of just telling the user in the README to be careful with casing, and maybe adding a similar warning to the error message, but that's just my preference; and as a user, I would consider that adequate/reasonable; and I wouldn't expect the maintainers to handle every possible casing scenario that could exist in a messy database.

leoebfolsom: @erezsh for sure--my only point was that I thought it was a good idea to give the user a heads up in the README that at least for Snowflake, table and schema names are case sensitive. Are you ok with the update I made in the README to that effect?

PR #124, opened 2022-06-28
reapply Presto and Snowflake support enhancements

sfenner-eb:
Snowflake:
Allow auth by PKCS8 key for Snowflake (password auth disabled)
User private key specification
Mapping bigint to Integer

Presto:
Allow verify http session with certificate for Presto
Allow using basic authentication for Presto
Presto Port specification
Presto HTTP scheme specification
Presto certificate specification
Mapping bigint to Integer

### DIFF
--- a/data_diff/databases/presto.py
+++ b/data_diff/databases/presto.py
@@ -30,14 +30,23 @@ class Presto(Database):
         "integer": Integer,
         "real": Float,
         "double": Float,
+        "bigint": Integer,
     }
     ROUNDS_ON_PREC_LOSS = True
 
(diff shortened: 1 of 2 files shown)


sfenner-eb:
RE: comment on last PR regarding imports

Shall I to project dependencies for these imports (they are included in Snowflake import for now, but not independently )
`from cryptography.hazmat.primitives import serialization
from cryptography.hazmat.backends import default_backend`

erezsh:
If we import that dependency (which your PR does), it's better to put the explicitly together with the snowflake connector. And to repeat my previous comment, they need to be imported "on-demand" inside `import_snowflake()`.

Issue #103, opened 2022-06-23
Let the differ choose a shared hashing algorithm


erezsh:
Right now we only support `md5` for hashing columns.

However, for some databases that might not be feasible. For example, in `mssql` it's too slow, and sounds like Spanner only supports sha1.

I think the solution should be to allow several implementations in subclasses of `Database`, and have the differ choose the best one that's shared between them. (or throw an error if there isn't one)

See issues #51 and #99.

sirupsen:
Agree. sha1 is also sometimes faster than md5, because some instruction-sets have specific instructions for it, but not for md5. crc32 is also generally faster than both of those.

abhinav1912:
I think this should be broken down into 2 tasks:
1. Adding a list of supported algorithms to the base/child classes, and refactor `md5_to_string` to something generic (maybe something like `hash_to_string`?)
2. Adding these changes to the interface/CLI where the DB type determines the intersection of hashing algorithms and offers a choice (and/or a default)

I'd be happy to start working on this, if it hasn't been picked up already.

sirupsen:
@abhinav1912 it'd be awesome if you could pick this up!

If you can do some research for (1) about what kinds of "alternate" negotiations we can do, and how you'd implement it, that'd be a great start before you start implementing it 👀 For example, for MSSQL, how can we do strings if we can't do md5 fast?

abhinav1912:
@sirupsen good point, I'll start looking into the alternatives and probably create a table of algorithms for each database. Will reach out on the Slack channel to clarify a few questions though.
As for MSSQL, [here's](http://kejser.org/exploring-hash-functions-in-sql-server/) a good article highlighting the performance of each hashing algorithm. `CHECKSUM` has good perfomance but poor hashing, `MD/SHA` algorithms have worse performance but better hashing. Doing better than `MD5` does seem a little difficult imo.

sirupsen:
That MD5 performance doesn't look that bad... I wonder why it came up so slow in our tests. Maybe the instance just didn't have enough CPU? Might be worth confirming that result first!

abhinav1912:
The performance point raises another question: is there be some script/utility that can calculate perf metrics for the current build? If not, then this could be a good addition.

PR #125, opened 2022-06-28
benchmark: add suite

sirupsen:
Cleanup of resolved merge conflict from https://github.com/datafold/data-diff/pull/67 based on all the changes in master. Trying to split it up into chunks to make it easier to review (and do an extra pass on it all)

The idea is to dual-purpose a subset of the type test-suite for benchmarking, to ensure it doesn't rust. In particular, to follow-up from https://github.com/datafold/data-diff/pull/106, the changes are to:

* In `BENCHMARK=1` mode, we do not want to drop tables, to prevent waiting for millions of rows to be inserted again. Not even across revisions if we're testing the performance change between two revisions
* Create indexes, to have good performance
* Write benchmark results to stdout and JSONL file(s) for graphing/analysis after
* Insert in larger batches for performance (follow-up on CSV)
* When benchmarking, we want to download segments in parallel for fairness

There are plenty of follow-ups to this, likely not all of which I will be able to complete:

* Enable threading > 1 in the test suite, which likely leads to some connection pooling logic changes I'd rather do parallel
* Do CSV insertion for the Cloud databases, to make it much faster to reproduce tests for 100M rows (currently takes hours)
* Add steps on how to reproduce the graph in the README based on `benchmark_*.jsonl`
* Add benchmarks with an increasing % of changed/deleted rows

@erezsh 

cc @abhinav1912 due to questions in https://github.com/datafold/data-diff/issues/103

### DIFF
--- a/data_diff/diff_tables.py
+++ b/data_diff/diff_tables.py
@@ -3,6 +3,7 @@
 
 from abc import ABC, abstractmethod
 import time
+import os
 from operator import attrgetter, methodcaller
 from collections import defaultdict
 from typing import List, Tuple, Iterator, Optional, Type
@@ -28,7 +29,7 @@ from .databases.database_types import (
 logger = logging.getLogger("diff_tables")
 
 RECOMMENDED_CHECKSUM_DURATION = 10
-
+BENCHMARK = os.environ.get("BENCHMARK", False)
 DEFAULT_BISECTION_THRESHOLD = 1024 * 16
 DEFAULT_BISECTION_FACTOR = 32
 
@@ -408,6 +409,16 @@ class TableDiffer:
             f"key-range: {table1.min_key}..{table2.max_key}, "
             f"size: {table2.max_key-table1.min_key}"
         )
+
+        # When benchmarking, we want the ability to skip checksumming. This
+        # allows us to download all rows for comparison in performance. By
+        # default, data-diff will checksum the section first (when it's below
+        # the threshold) and _then_ download it.
+        if BENCHMARK:
+            max_rows_from_keys = max(table1.max_key - table1.min_key, table2.max_key - table2.min_key)
+            if max_rows_from_keys < self.bisection_threshold:
+                yield from self._bisect_and_diff_tables(table1, table2, level=level, max_rows=max_rows_from_keys)
+                return
 
         (count1, checksum1), (count2, checksum2) = self._threaded_call("count_and_checksum", [table1, table2])
 
--- a/tests/test_database_types.py
+++ b/tests/test_database_types.py
@@ -1,16 +1,19 @@
 from contextlib import suppress
 import unittest
 import time
+import json
 import re
+import rich.progress
 import math
 import uuid
 from datetime import datetime, timedelta
+import logging
 from decimal import Decimal
 from parameterized import parameterized
 
 from data_diff import databases as db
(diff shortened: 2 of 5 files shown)


data_diff/diff_tables.py
```diff
@@ -403,6 +404,15 @@ def _diff_tables(self, table1, table2, level=0, segment_index=None, segment_coun
             f"size: {table2.max_key-table1.min_key}"
         )
 
+        # When benchmarking, we want the ability to download the segments in
+        # parallel. By default, data-diff will checksum the section first (when
+        # it's below the threshold) and _then_ download it.
+        if BENCHMARK:
```
sirupsen, line 410: FWIW I don't think it would be crazy to make this the default... 

I ran the benchmark with the default bisection threshold, and checksumming / downloading with the default threshold and ~10k rows is extremely competitive. And the current approach is twice as slow when there is a difference.

data_diff/diff_tables.py
```diff
@@ -403,6 +404,15 @@ def _diff_tables(self, table1, table2, level=0, segment_index=None, segment_coun
             f"size: {table2.max_key-table1.min_key}"
         )
 
+        # When benchmarking, we want the ability to download the segments in
```
erezsh, line 407: I don't understand the word "parallel" in this context.

Wouldn't it be more correct to say "we want the ability to skip the checksum, and force the download of all segments" ?

tests/test_database_types.py
```diff
@@ -356,33 +359,90 @@ def expand_params(testcase_func, param_num, param):
...
 
-def _insert_to_table(conn, table, values):
-    insertion_query = f"INSERT INTO {table} (id, col) "
+def _insert_to_table(conn, table, values, type):
+    current_n_rows = conn.query(f"SELECT COUNT(*) FROM {table}", int)
+    if current_n_rows == N_SAMPLES:
+        assert BENCHMARK, "Table should've been deleted, or we should be in BENCHMARK mode"
+        return
+    elif current_n_rows > 0:
+        logging.error(f"{conn.name}: Table {table} has {current_n_rows}, expected {N_SAMPLES} or 0. Recreating table")
```
erezsh, line 368: Maybe this should be a warning?

sirupsen: Default logging error level is ERROR to have less noise for test output.

data_diff/diff_tables.py
sirupsen: Sure.

sirupsen:
> mysql.connector.errors.ProgrammingError: 1059 (42000): Identifier name 'idx_dst_postgresql_double_precision_to_mysql_numeric65_10_50_id_col' is too long

Ugh, not sure how I missed that! Fixed with another `.replace`

sirupsen:
@erezsh tried to resolve your comments! Let me know if it's good to go now :)

tests/test_database_types.py
erezsh: Maybe I'm missing something, but it should be simple to set the test logger to warning, and the data-diff loggers to error, no?

sirupsen: Yeah, that was what I was getting at in my last setntence.. I will just remove tit then

sirupsen:
Good to merge?

PR #128, opened 2022-06-29
doc: list of hashing algorithms supported by DBs

abhinav1912:
Added table containing all hashing algorithm choices. Related to issue #103 
cc: @sirupsen

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,18 @@ $ data-diff \
 If a database is not on the list, we'd still love to support it. Open an issue
 to discuss it.
 
+##### Hashing Algorithms supported by databases
+
+| Database      | Supported algorithms                        | Resources                                        |
+|---------------|---------------------------------------------|--------------------------------------------------|
+| PostgreSQL    | MD5, SHA224, SHA256, SHA384, SHA512         |  [Doc][PSQL_Doc]                                 |
+| MSSQL         | MD2, MD4, MD5, SHA, SHA1                    |  [Perf][MSSQL_Perf]                              |
+| Snowflake     | Hash/Hash_Agg, MD5, SHA1, SHA2              |  [Doc][Snowflake_Doc]                            |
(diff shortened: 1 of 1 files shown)


erezsh:
Thanks for collecting this information. Please put it in [Discussions](https://github.com/datafold/data-diff/discussions).

We will not add to the README a list of algorithms that we don't support.

sirupsen:
@abhinav1912 this is an excellent collection, thank you! 😍 

As Erez pointed out, I think we should keep it in the issue at-hand, as it won't be _too_ useful for the general user I think.

erezsh:
I think that since the only issue here is performance, and the only one in mssql that performs well is not shared by anyone else.. the only path forward I can see is to implement some hash function using simple arithmetic, and try to prove that it's good enough to be used for comparing data.

sirupsen:
Yeah, I think the next step is to target MSSQL directly (which I assume you have experience with based on your existing posts?), we can look at others (e.g. Spanner) later.

I think we should investigate a `sum-number` as the first alternative to `md5`. Spanner is the only one I've seen so far that supports hashing, just not MD5.

1) What is the MD5 performance in MSSQL compared to Postgres for 1M records for e.g. an int column?

2) If it's really bad, what alternative scheme can we come up with for MSSQL? 

For `int`, `float`, and `datetime` we could do a `sum()` (which is also what we'd negotiate with e.g. ElasticSearch). I think negotiating a `sum-number` algorithm could be a good start, since for every database this will also be much faster than MD5, so it's useful regardless.

3) Can we do `sum` on a string? Is there a standard way to get e.g. a sum of all the UTF-8 characters?

abhinav1912:
@sirupsen of the hashing algorithms, `MD5` is the best option. We could try using either `CHECKSUM` or `BINARY_CHECKSUM` (with some key points highlighted [here](https://stackoverflow.com/questions/1240425/what-are-the-differences-between-checksum-and-binary-checksum-and-when-what)), since both of these also work with strings.

PR #135, opened 2022-06-30
tests: test snowflake in github actions

sirupsen:
Add Snowflake and Presto to CI with `unittest-parallel`, which required a bunch of changes:

1. Remove the initial type tests in favour of only `test_database_types.py`, since Preql doesn't support e.g. Presto. This meant moving some timezone tests, but I think this simplifies things long-term!
2. Ensure that all tests are safe to be run concurrently both with the same suite running in parallel, or other tests. Every test's tables now get unique names.
3. Add secrets for Snowflake and Presto which can be set through the environment in Github actions.
4. Tests aren't run for first-time collaborators by default. This is to prevent people from being able to fork, put up a PR, and leak our secrets. It's only for first-time collaborators it is required though, so for repeat collaborators, they won't have issues.

Redshift, BigQuery, and Oracle are still missing. They require some more configuration.


erezsh (approved):
👍

PR #171, opened 2022-07-22
Now separating the columns with a '|' before concatting them

erezsh:


erezsh:
@nolar Does that answer your concerns?

nolar (approved):

Issue #150, opened 2022-07-05
Dump the config for debugging but with secrets redacted


nolar:
_As a follow-up for a discussion at https://github.com/datafold/data-diff/pull/143#discussion_r913607678._

**Why?** It is helpful for debugging when the users show their logs with their configs. However, configs contain database passwords, which should be printed. 

**Problem 1:** Printing the configs with passwords to stdout/stderr is considered a security leak in some organizations, which requires changing the leaked passwords and reporting a security incident — e.g. if data-diff is executed in a remote environment (Docker containers, Kubernetes Jobs/Pods, etc).

**Problem 2:** With the realistic database tests from #112, the passwords should have been defined as GitHub Actions Sections. GitHub automatically redacts the secrets if/when it can detect them, but still recommends that we do not print them in the first place:

* https://docs.github.com/en/actions/security-guides/encrypted-secrets#accessing-your-secrets

**Goal:**

* Print/log the used config with all secrets redacted both from the config keys and from URI.

**Notes:**

A malicious contributor can simply `print()` them (with modifications to work around GitHub's redacting) — but GitHub usually requires clicking a button for external (non-trusted) contributors before running the Actions — at least some protection.

PR #172, opened 2022-07-22
Print configuration during debug, but with passwords redacted

erezsh:
Solves issue #150

### DIFF
--- a/data_diff/utils.py
+++ b/data_diff/utils.py
@@ -54,7 +54,18 @@ def number_to_human(n):
     return "{:.0f}{}".format(n / 10 ** (3 * millidx), millnames[millidx])
 
 
-def remove_password_from_url(url, replace_with="***"):
+def _join_if_any(sym, args):
+    args = list(args)
+    if not args:
+        return ''
+    return sym.join(str(a) for a in args if a)
+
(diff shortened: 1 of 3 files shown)


data_diff/utils.py
```diff
@@ -51,3 +52,9 @@ def number_to_human(n):
     )
 
     return "{:.0f}{}".format(n / 10 ** (3 * millidx), millnames[millidx])
+
+
+def remove_password_from_url(url, replace_with="***"):
+    parsed = urlparse(url)
+    replaced = parsed._replace(netloc="{}:{}@{}".format(parsed.username, replace_with, parsed.hostname or ""))
```
nolar, line 59: **Important 1:** It loses the port value here if it is set, which is often essential for users.

**Important 2:** It might inject the `***` even if no password was there at all.

Similarly, it will inject the literal string `None` in place of a username if no username was specified there originally, or maybe it will insert an unnecessary `:@` prefix — e.g. when using different auth methods than the "username:password" pair; maybe even with tokens.

erezsh: You're right, I can do it better. thanks!

nolar (approved):

PR #178, opened 2022-07-26
Remove numpy dependencies <1.22.0 due to security issue with…

cfernhout:
…numpy.core component. CVE-2021-34141

`snowflake-connector-python` latest version bumped dependency version of NumPy from <1.23.0 to <1.24.0. Although it is a pretty harmless security issue, there is an easy fix.


cfernhout:
Unfortunately, it is conflicting with 3.7 support. Numpy 1.22.0 and onward requires Python 3.8 or higher.

nolar:
Then I suggest we keep it as is, i.e. not restricted, and let the users update their final app environments accordingly.

Anyway, the CVE mentions "constructing specific string objects", but data-diff does not accept input from arbitrary users. And it does not use numpy directly or indirectly — that is probably some form of ["dataframe storing/fetching"](https://docs.snowflake.com/en/user-guide/python-connector-pandas.html) in the connector, which we do not do. 

So it should be safe.

Issue #59, opened 2022-06-14
Non-integer IDs


joeltone:
I have several tables that use GUIDs for their primary key.  Can data-diff test these tables or only tables where the primary key is an integer?

sirupsen:
@joeltone are you using the UUID type in Postgres? What's source/target database and type?

joeltone:
Both databases are in Snowflake and I'm just using a string data type for the ids.

joeltone:
I'll see what I can find.  Thanks for the help!

sirupsen:
Below's a really hack prototype of how it's possible to support string keys, assuming a random distribution (as UUIDs would have). Probably has some crucial bugs and off-by-ones, but I think directionally it'd work.

```patch
diff --git a/data_diff/database.py b/data_diff/database.py
index dc208b6..cafeb5f 100644
--- a/data_diff/database.py
+++ b/data_diff/database.py
@@ -216,7 +216,7 @@ class Database(AbstractDatabase):
         rows = self.query(self.select_table_schema(path), list)
 
         # Return a dict of form {name: type} after canonizaation
-        return {row[0].lower(): self._parse_type(*row[1:]) for row in rows}
+        return {row[0]: self._parse_type(*row[1:]) for row in rows}
 
     @lru_cache()
     def get_table_schema(self, path: DbPath) -> Dict[str, ColType]:
diff --git a/data_diff/diff_tables.py b/data_diff/diff_tables.py
index e25bf15..c25b7f2 100644
--- a/data_diff/diff_tables.py
+++ b/data_diff/diff_tables.py
@@ -28,8 +28,18 @@ def safezip(*args):
 
 
 def split_space(start, end, count):
-    size = end - start
-    return list(range(start, end, (size + 1) // (count + 1)))[1 : count + 1]
+    space = []
+    for c in range(ord('0'), ord('9')):
+        space.append(start + chr(c))
+
+    for c in range(ord('a'), ord('f')):
+        space.append(start + chr(c))
+
+    print(space)
+    return space
+
+    # size = end - start
+    # return list(range(start, end, (size + 1) // (count + 1)))[1 : count + 1]
 
 
 def parse_table_name(t):
@@ -96,9 +106,9 @@ class TableSegment:
 
     def _make_key_range(self):
         if self.min_key is not None:
-            yield Compare("<=", str(self.min_key), self._key_column)
+            yield Compare("<=", f"'{self.min_key}'", self._key_column)
         if self.max_key is not None:
[...]

PR #179, opened 2022-07-26
Initial support for Alphanumeric IDs + tests (Issue #59)

erezsh:


nolar (approved):
A little concern: will it handle UUIDs stored or returned as strings with dashes — i.e. `1234-5678ABCD-………-98765432`.

I presume it will not be able to handle the non-alphanumeric strings with the extended alphabet (dots, pluses, dashes, etc.), so as alphanum strings with mixed-case alphabet (`aBcD`, which is different from `abcd`) — and this is intended, right?

erezsh:
UUIDs are handled separately.

Strings with arbitrary dashes are not supported. "Extended alphabet" (non-alphanums) also not supported.

Mixed-case is supported. However the keys "abc" and "ABC" will be considered as the same key, which we may want to change in the future.

PR #181, opened 2022-07-27
Update README.md

kning:
Postgres key is `host` not `hostname` see https://github.com/datafold/data-diff/blob/master/data_diff/databases/postgresql.py#L53


erezsh:
Thank you for bringing it to our attention.

Can you please link to where it says that in the docs? (or anywhere official)

kning:
It came up when I was running example locally, I got this error:

image

Looks it's just using psycopg2 library to open the connection, which is just `host`: https://www.psycopg.org/docs/module.html

erezsh:
Thanks!

Issue #83, opened 2022-06-21
Add --all-mutual-columns


sirupsen:
In some cases, you might want to checksum _everything_, and instead of having to type out all the columns yourself, we should be able to get them from the schema.

`--all-mutual-tables` could be a nice addition at some point too...

erezsh:
Implemented with the new `-c pattern` feature.

Instead of `--all-mutual-tables`, users can write `-c %`.

PR #185, opened 2022-07-28
Added support for auto-detecting mutual columns, and using patterns in -c

erezsh:
Solves issue #83 .

@nolar This one didn't turn out as pretty as I'd like it to be.

Any suggestions? Or maybe it's the least of all evils?

**Explanation:**

We don't want to process column types that the user didn't request, because they might generate irrelevant warnings, or even errors for unsupported types. (and we also query the table for them, to figure out their 'refined' type)

But, we need the schema in order to choose the mutual columns.

And we need the mutual columns to choose which columns to process.

This is cyclical, so I break the cycle by working with "raw" schema until the mutual columns can be figured out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,7 +144,7 @@ If a database is not on the list, we'd still love to support it. Open an issue
 to discuss it.
 
 Note: Because URLs allow many special characters, and may collide with the syntax of your command-line,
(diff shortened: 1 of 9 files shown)


erezsh:
(waiting to test on databricks)

PR #200, opened 2022-08-10
New database driver: Exasol

nklsw:
I'm happy to introduce a **new db-driver** for [Exasol](https://www.exasol.com/)

#### Local Dev Setup
1. Bring up the Exasol Docker container:
```
docker-compose up -d exasol
```

2. Download Sample Data
```
curl https://datafold-public.s3.us-west-2.amazonaws.com/1m.csv -o dev/ratings.csv
```

3. Run python script to import CSV data into Exasol
```
poetry run python3 dev/exasol/_exasol_import_csv.py
```

4. Run data-diff
```
poetry run python3 -m data_diff exasol://sys:exasol@localhost/DATADIFF rating exasol://sys:exasol@localhost/DATADIFF rating_del1 --verbose
```

#### Unittest
1. Bring up mysql and postgres 
```
docker-compose up -d mysql postgres
```
2. Set env variable for Exasol URI
```
export DATADIFF_EXASOL_URI=exasol://sys:exasol@localhost/DATADIFF
``` 
4. Run unittests
```
poetry run unittest-parallel -j 16
```

_Note on Exasol Docker: There is currently an [issue](https://github.com/exasol/docker-db/issues/45) with Exasol and docker-compose which could lead to errors when restarting the existing container. If this happens you need to run the script in `./dev/exasol/reset_ip.sh` to which will reset the IP config for Exasol and fix the problem._


nklsw:
@erezsh I fixed the numpy version for Python 3.7.1 in poetry.lock and cleaned up my commits.

data_diff/databases/exasol.py
```diff
@@ -0,0 +1,161 @@
...
+        if not rows:
+            raise RuntimeError(f"{self.name}: Table '{'.'.join(path)}' does not exist, or has no columns")
+
+        d = {r[0]: r for r in rows}
+        assert len(d) == len(rows)
+        return d
+
+    def _process_table_schema(self, path: DbPath, raw_schema: Dict[str, tuple], filter_columns: Sequence[str]):
+        accept = {i.lower() for i in filter_columns}
+        print('##DEBUG##')
```
erezsh: You forgot to remove this.

data_diff/databases/exasol.py
```diff
@@ -0,0 +1,161 @@
...
+    def query_table_schema(self, path: DbPath) -> Dict[str, tuple]:
+        rows = self.query(self.select_table_schema(path), list)
+        if not rows:
+            raise RuntimeError(f"{self.name}: Table '{'.'.join(path)}' does not exist, or has no columns")
+
+        d = {r[0]: r for r in rows}
+        assert len(d) == len(rows)
+        return d
+
+    def _process_table_schema(self, path: DbPath, raw_schema: Dict[str, tuple], filter_columns: Sequence[str]):
```
erezsh, line 93: Why did you override `_process_table_schema()`? It seems the same as the original.

data_diff/databases/exasol.py
```diff
@@ -0,0 +1,161 @@
...
+        return _query_conn(self.thread_local.conn, sql_code)
+
+    def select_table_schema(self, path: DbPath) -> str:
+        schema, table = self._normalize_table_path(path)
+
+        return (
+            f'SELECT "COLUMN_NAME", "COLUMN_TYPE", 3 AS \'datetime_precision\', "COLUMN_NUM_PREC", "COLUMN_NUM_SCALE" FROM SYS.EXA_ALL_COLUMNS WHERE "COLUMN_TABLE" = \'{ table }\' AND "COLUMN_SCHEMA" = \'{ schema }\''
+        )
+
+    def query_table_schema(self, path: DbPath) -> Dict[str, tuple]:
```
erezsh: Why did you override `query_table_schema()`? It seems the same as the original.

data_diff/databases/exasol.py
```diff
@@ -0,0 +1,161 @@
...
+            return f"RPAD({truncated}, {TIMESTAMP_PRECISION_POS+6}, '0')"
+
+    def normalize_number(self, value: str, coltype: FractionalType) -> str:
+        return self.to_string(f"cast({value} as DECIMAL(18, {coltype.precision}))")
+
+    def _query(self, sql_code: str):
+        r = self._queue.submit(self._query_in_worker, sql_code)
+        return r.result()
+
+    def _query_in_worker(self, sql_code: str):
```
erezsh: This is the same as the original, no need to override.

data_diff/databases/exasol.py
```diff
@@ -0,0 +1,161 @@
...
+        super().__init__(thread_count=thread_count)
+
+    def create_connection(self):
+        self._exasol = import_exasol()
+        try:
+            c = self._exasol.connect(**self.kwargs)
+            if SESSION_TIME_ZONE:
+                c.execute(f"ALTER SESSION SET TIME_ZONE = '{SESSION_TIME_ZONE}'")
+            return c
+        except Exception as e:
```
erezsh, line 44: Please catch an exasol base exception here, not just every exception.

data_diff/databases/exasol.py
```diff
@@ -0,0 +1,161 @@
...
+            raise ConnectError(*e.args) from e
+
+    def _query(self, sql_code: str):
+        try:
+            return super()._query(sql_code)
+        except Exception as e:
```
erezsh, line 51: Same here.

erezsh (changes requested):
Please fix the indicated issues with the code.

Duplicated methods and unneeded overrides will make future maintenance harder. Catching all exceptions can hide bugs in your implementation, and make it harder to debug.

Also, please run `black -l 120` on all the files, because it's the decided formatting standard for the project. (and makes it easier to review the code)

data_diff/databases/exasol.py
nklsw: I need to override this to return my customized [`_query_conn`](https://github.com/nklsw/data-diff/blob/a9c7a87bf1e1447c98b3defb7246d78fe4bb9a6f/data_diff/databases/exasol.py#L17) because the method is not a member of the base class.

data_diff/databases/exasol.py
nklsw: Already removed this in ffaf4a8ab3c9118fa60578f0f8485579812e44ad

data_diff/databases/exasol.py
nklsw: Already removed this in ffaf4a8ab3c9118fa60578f0f8485579812e44ad

data_diff/databases/exasol.py
nklsw: Fixed in a5c90dd

data_diff/databases/exasol.py
nklsw: Fixed in a5c90dd

data_diff/databases/exasol.py
erezsh: Oh, I understand. In that case you can leave it. We might later add `_query_conn` as a method and change it.

Btw, does Exasol not support creating a cursor?

data_diff/databases/exasol.py
nklsw: I somehow removed some arguments for _parse_type() which led to the modified _process_table_schema().
But actually, there is no need to do this. I fixed both methods in [0c1384b](https://github.com/datafold/data-diff/pull/200/commits/0c1384b380b1ee2621049bd9030badb9ed6ee371)

nklsw:
@erezsh Thanks for your indications! I fixed the issues and ran black on all files. 

Please have a look at the remaining [open discussion](https://github.com/datafold/data-diff/pull/200#discussion_r943478280).

data_diff/databases/exasol.py
nklsw: Okay 👍

No, Exasol does not support cursors.

erezsh:
@nklsw The tests fail unless I import the csv, since the schema needs to be created. But also calling "create schema" fails with a schema not found error. The only way forward I can see is to connect without a schema (like in the import) and create the schema, before doing `connect_to_uri()`, but then we can't use the URI, which means the user will have to specify another env variable for a URI without a schema, which is a little awkward.

Can you think of an elegant solution to this problem?

erezsh:
I ended up adding this:

```python
if CONNS[db.Exasol]:
    kw = dict(CONNS[db.Exasol].kwargs)
    kw['driver'] = 'exasol'
    kw['host'] = kw.pop('dsn')
    schema = kw.pop('schema')
    conn_without_schema = db.connect.connect(kw)
    conn_without_schema.query(f'CREATE SCHEMA IF NOT EXISTS {schema}', None)
```

erezsh:
I fixed everything, see PR #203 

The only problem left that I wasn't sure how to fix are these lines:

```python
        if target_db is db.Exasol and source_db is db.PostgreSQL and source_type == "timestamp with time zone":
            values_in_source = ((a, b.replace(tzinfo=None)) for a, b in values_in_source)
```

This doesn't look right. Maybe you need to set the timezone to 0 in Exasol itself?

nklsw:
The problem is, that `_insert_to_table()` would try to insert something like `2022-08-12 12:23:45+0000` but Exasol only accepts `YYYY-MM-DD HH24:MI:SS.FF6` for inserting timestamps. I'm currently not sure, how to deal with it in a better way, but I agree with you, this doesn't look robust now.

erezsh:
@nklsw So the problem is just with the format of the date? Then you can modify `_insert_to_table()` to reformat it according to the target db.

PR #213, opened 2022-08-22
Added optional tracking

erezsh:


erezsh:
The two events as seen on freshpaint - 

image

data_diff/diff_tables.py
```diff
@@ -322,45 +330,79 @@ def diff_tables(self, table1: TableSegment, table2: TableSegment) -> DiffResult:
...
-        if not isinstance(key_type, IKey):
-            raise NotImplementedError(f"Cannot use column of type {key_type} as a key")
-        if not isinstance(key_type2, IKey):
-            raise NotImplementedError(f"Cannot use column of type {key_type2} as a key")
-        assert key_type.python_type is key_type2.python_type
-
-        # Query min/max values
-        key_ranges = self._threaded_call_as_completed("query_key_range", [table1, table2])
+        self.stats["diff_count"] = 0
+        start = time.time()
```
nolar, line 339: **nit:** `time.time()` is not monotonic, it can go backward or jump forward, thus distorting the measurements — more likely on local time/clock adjustments by ntp or by the user.

For performance measurements, `time.monotonic()` or `time.perf_counter()` are used. For data-diff's scale, the first one is sufficient (the second one has higher precision, which is not needed).

However, in the majority of cases, `time.time()` will work sufficiently good too.

data_diff/tracking.py
```diff
@@ -0,0 +1,114 @@
...
+    try:
+        with open(DEFAULT_PROFILE) as f:
+            conf = toml.load(f)
+    except FileNotFoundError as e:
+        conf = {}
+
+    if "anonymous_id" not in conf:
+        conf["anonymous_id"] = str(uuid4())
+        with open(DEFAULT_PROFILE, "w") as f:
+            toml.dump(conf, f)
```
nolar, line 33: Just to clarify: this file is not used for anything but for datadiff's internal state, correct?

(Why? => This operation will reformat the config and will remove all comments from it. If the config is crafted by humans, they will be unhappy. If it is machine-generated, there is no problem.)

data_diff/tracking.py
```diff
@@ -0,0 +1,114 @@
...
+    return g_anonymous_id
+
+
+def create_start_event_json(diff_options: Dict[str, Any]):
+    return {
+        "event": "os_diff_run_start",
+        "properties": {
+            "distinct_id": get_anonymous_id(),
+            "token": TOKEN,
+            "time": time(),
```
nolar, line 64: Contrary to what I said about `time.monotonic()` above, this is a good place for `time.time()` — i.e. the UTC moment when it was executed.

data_diff/tracking.py
```diff
@@ -0,0 +1,114 @@
...
+            "table_2_rows_cnt": table2_count,
+            "diff_rows_cnt": diff_count,
+            "error_message": error,
+        },
+    }
+
+
+def send_event_json(event_json):
+    if not g_tracking_enabled:
+        raise RuntimeError("Won't send; tracking is disabled!")
```
nolar, line 103: Is this an error? Shouldn't it silently exit instead — as per users' expectations?

nolar (approved):

data_diff/tracking.py
erezsh: Currently it is only used by the machine. I was thinking of making it a "global" config, in a future PR.

If you have an idea on how to preserve human-happiness artifacts, I'll be happy to include it.

data_diff/tracking.py
erezsh: `send_event_json()` should never run if tracking is disabled. This is just another layer of protection, to prevent human error.

nolar: Ah, I see. Got it. THanks.

data_diff/tracking.py
nolar: There is no way to preserve that. I would suggest using another file with the anonymous id in that case (e.g. `~/.datadiff-id.toml`).

PR #217, opened 2022-08-26
Support of Clickhouse

pik94:


pyproject.toml
```diff
@@ -23,7 +23,7 @@ classifiers = [
 packages = [{ include = "data_diff" }]
 
 [tool.poetry.dependencies]
-python = "^3.7.1"
+python = "^3.8.0"
```
nolar, line 26: Do we intentionally drop Python 3.7 while it is still officially maintained (until June 2023)?

pik94: No, we do not. It just was for testing because updating the lock with python 3.7 was x100 slower than for python 3.8.
New changes reverts python 3.7 back.

data_diff/databases/clickhouse.py
```diff
@@ -0,0 +1,153 @@
+from typing import Optional, Type
+
+import clickhouse_driver.dbapi.connection
```
erezsh, line 3: Shouldn't have this import outside of `import_clichouse()`

data_diff/databases/clickhouse.py
```diff
@@ -0,0 +1,153 @@
...
+        try:
+            connection = SingleConnection(**self._args)
+            return connection
+        except clickhouse.OperationError as e:
+            raise ConnectError(*e.args) from e
+
+    def _parse_type_repr(self, type_repr: str) -> Optional[Type[ColType]]:
+        nullable_prefix = 'Nullable'
+        if type_repr.lower().startswith(nullable_prefix.lower()):
+            type_repr = type_repr[len(nullable_prefix):].lstrip('(').rstrip(')')
```
erezsh, line 75: This isn't readable. Please use a regexp.

data_diff/databases/clickhouse.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def to_string(self, s: str) -> str:
+        return f"toString({s})"
+
+    def normalize_timestamp(self, value: str, coltype: TemporalType) -> str:
+        if coltype.rounds:
+            prec = coltype.precision
+            timestamp = f'toDateTime64(round(toUnixTimestamp64Micro(toDateTime64({value}, 6)) / 1000000, {prec}), 6)'
+            return self.to_string(timestamp)
+        else:
```
erezsh, line 101: "else" clause not needed. I suggest using pylint on this module to catch things like that.

data_diff/databases/clickhouse.py
```diff
@@ -0,0 +1,153 @@
...
+            fractional = f'toUnixTimestamp64Micro(toDateTime64({value}, 6)) % 1000000'
+            return f"formatDateTime({value}, '%Y-%m-%d %H:%M:%S') || '.' || {self.to_string(fractional)}"
+
+    def _convert_db_precision_to_digits(self, p: int) -> int:
+        # Done the same as for PostgreSQL but need to rewrite in another way
+        # because it does not help for float with a big integer part.
```
erezsh, line 107: I don't understand this comment. Can you explain it? What floats, for example, it doesn't help with?

data_diff/databases/clickhouse.py
```diff
@@ -0,0 +1,153 @@
...
+        # If a decimal value has trailing zeros in a fractional part, when casting to string they are dropped.
+        # For example:
+        #   select toString(toDecimal128(1.10, 2));  -- the result is 1.1
+        #   select toString(toDecimal128(1.00, 2)); -- the result is 1
+        # So, we should use some custom approach to save these trailing zeros.
+        # To avoid it, we can add a small value like 0.000001 to prevent dropping of zeros from the end when casting.
+        # For examples above it looks like:
+        #   select toString(toDecimal128(1.10, 2 + 1) + toDecimal128(0.001, 3)); -- the result is 1.101
+        # After that, cut an extra symbol from the string, i.e. 1.101 -> 1.10
+        # So, the algorithm is:
```
erezsh, line 120: Why is this better than calling RPAD, once with "." and once with "0" ?

It sounds like your method might lead to all kinds of strange precision errors.

data_diff/databases/clickhouse.py
```diff
@@ -0,0 +1,153 @@
...
+        'UUID': Native_UUID,
+
+        'String': Text,
+        'FixedString': Text,
+
+        'DateTime': Timestamp,
+        'DateTime64': Timestamp,
+
+    }
+    ROUNDS_ON_PREC_LOSS = True
```
erezsh, line 55: In the tests you wrote "Clickhouse does not round microseconds when inserting in contrast to PostgreSQL and MySQL.", so sounds like `ROUNDS_ON_PREC_LOSS ` should be False.

erezsh (changes requested):
Also, run `black -l 120 data_diff`, as mentioned here - https://github.com/datafold/data-diff/blob/master/CONTRIBUTING.md#code-style

data_diff/databases/clickhouse.py
```diff
@@ -0,0 +1,153 @@
...
+        except clickhouse.OperationError as e:
+            raise ConnectError(*e.args) from e
+
+    def _parse_type_repr(self, type_repr: str) -> Optional[Type[ColType]]:
+        nullable_prefix = 'Nullable'
+        if type_repr.lower().startswith(nullable_prefix.lower()):
+            type_repr = type_repr[len(nullable_prefix):].lstrip('(').rstrip(')')
+
+        if type_repr.startswith('Decimal'):
+            type_repr = 'Decimal'
```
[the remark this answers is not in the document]
pik94: > Aren't we throwing valuable information here, like precision/scale?

No, we are not. This information comes from `information_schema` anyway.

data_diff/databases/clickhouse.py
pik94: > Why is this better than calling RPAD, once with "." and once with "0" ?

It's a problem to obtain a number of digits (it may vary in contrast to timestamp format) to apply `rpad` here.

> It sounds like your method might lead to all kinds of strange precision errors.
Does it? I have just added a small number with a higher precision here, and do not use it to round something.

.github/workflows/ci.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         python-version:
-          - "3.7.1"
+          - "3.7"
```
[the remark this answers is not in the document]
erezsh: Oh, I see. I thought we had it at 3.7.1 for a different reason, but if that's the case then it's fine.

Btw, what did you change to make updating the lockfile work?

data_diff/databases/clickhouse.py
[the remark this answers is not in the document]
erezsh: But `type_repr` *is* the information that comes from `information_schema`, no?

data_diff/databases/clickhouse.py
erezsh: I can't say for sure. It just feels unstable, but maybe I'm worried for nothing.

I'll give it a deeper look later.

.github/workflows/ci.yml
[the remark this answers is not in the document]
pik94: Nothing special, it just updated without a resolving error although it failed with it before. For me it'a a kind of magic how poetry works...

data_diff/databases/clickhouse.py
pik94: It's dirty hack to save a compatibility with Postgres. 
Clickhouse `float32` use `float` of C language under the hood. I guess C float is used for Postgres' 4-byte `real` (or `float4` that is the same) too because it has the same behavior. It still remains a mystery to me why 2 is [subtracted](https://github.com/datafold/data-diff/blob/master/data_diff/databases/postgresql.py#L49) here for PostgreSQL. I think it's not wierd precision issues as mentioned in a comment for postgres but a normal behavior of 4-byte floating number which has 6 precision. For float it means it can store no more than 6-7 digits in average. Let me clarify on example for Postgres (the same fot Clickhouse):
```
with tmp as (
    select 0.123456789::float4 x
    union
    select 10.123456789::float4 x
    union
    select 100.123456789::float4 x
    union
    select 1000.123456789::float4 x
)
select
    x::decimal(38, 6)
from tmp
order by x;
```

Result:
```
      x      
-------------
    0.123457
   10.123500
  100.123000
 1000.120000
(4 rows)
```
If we increase an integer part 10 times, we leave 1 digit of a floating part. So, 6-7 precision for float means that it can store 6-7 digits for whole number, namely, integer + float parts. 

I see in unit tests that  we have a number [100.98](https://github.com/datafold/data-diff/blob/master/tests/test_database_types.py#L327), and I think tests are passed because of substracting of 2 from precision for Postgres. But if you add a number with a higher integer part, for example, 1000.98, tests failed.

My opinion, we cannot cast `float` to `decimal(38, some_precision)` because precision here depends on the integer part of a float value.

erezsh: It does look like the precision of floats is a little more complicated than we currently account for.

But following your example, I can think of another problem:

```sql
with tmp as (
    select 123456789::float4 x                                                                                                                                                     )
select
    x::decimal(38, 6)
from tmp;

        x
------------------
 123457000.000000
```

So even the integer parts aren't safe!

I do think we can make two assumptions that helps us a lot -

1) The value already sits in some database, and probably was given a datatype that makes sense on both ends

2) We only need to be as precise at the tool that copied it (probably in a migration).

But yeah, I agree this is a weak point, and it deserves some more attention.

data_diff/databases/clickhouse.py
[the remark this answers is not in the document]
erezsh: I understand. Ideally I would like to read it anyway, and just add an assert to make sure it's the same as the `information_schema`. But I won't insist on it.

data_diff/databases/clickhouse.py
masonwheeler: @pik94 It's not "C language `float`s," it's the IEEE 754 floating-point standard that defines what `float` should look like and how it should behave for all digital computers in order to ensure interoperability.  C (and virtually every other programming language) does it that way because the CPU does it that way.

pik94: @masonwheeler, yes, it's more accurate to say that it's behavior of the floating point number described in the IEEE 754 standard. I just used `float` because it's mentioned in clickhouse [docs](https://clickhouse.com/docs/en/sql-reference/data-types/float/).

data_diff/databases/clickhouse.py
pik94: To the current moment I see only one corner case, namely, if a customer's number type has precision higher than 38 digits (for example, they can use `Decimal256`, or `Decimal(100, 40)` etc). Casting to `Decimal128(39)`is not possible. However, I'm note sure that people has such high precision.

Also I tried to rewrite it using RPAD but found with next problem. When we do not add a little value, casting of float to decimal has an unpleasant side effect. For example, let's we want to cast to precision equaled 5:
```
with tmp as (
    select toFloat64(0.001201923076923077) as col
)
select
    col,
    round(col, 5),
    toDecimal128(round(col, 5), 5) as precision_5,
    toDecimal128(round(col, 5), 6) as precision_6
from tmp
order by col;
```
Result:
```
0.001201923076923077, 0.0012, 0.00119, 0.001200
```
As you can see, the number was corrupted for precision for 5.

data_diff/databases/clickhouse.py
pik94: Done.

data_diff/databases/clickhouse.py
pik94: Done.

data_diff/databases/clickhouse.py
pik94: Rewrote it in a more readable form but without regexp.

data_diff/databases/clickhouse.py
pik94: Done.

erezsh:
You took the snowflake connector too far back, it doesn't work for Python 3.10 - https://github.com/datafold/data-diff/pull/226

(the unittests in this PR didn't run snowflake)

pik94:
> You took the snowflake connector too far back, it doesn't work for Python 3.10 - https://github.com/datafold/data-diff/runs/8187675999?check_suite_focus=true
> 
> (the unittests in this PR didn't run snowflake)

@erezsh hm... Is there a way to run it here

erezsh:
Let's worry about that next time :)

PR #219, opened 2022-08-29
Adds logger 

gaurav1308:


erezsh:
This is for you to run locally and let me know what's the result. Because only you have access to your database, that's causing the code to fail.

It doesn't belong in the official code-base.

PR #236, opened 2022-09-15
Query builder + Join-diff (in-db)

erezsh:
Big PR containing -

- Query builder package, for easy construction of SQL queries with Python. While not ready yet for all use-cases, it's complete enough for the data-diff use-case.
- Join-diff (in-db) implementation of data-diff, with collection of extra statistics  (tracking not yet implemented)
- Existing differ moved to "hashdiff_tables" module, and updated to use the new query builder
- Updated the CLI and Python API to allow using both algorithms.

New CLI interface works as follow:

```
data-diff db1 table1 db2 table2    # hashdiff (just as before)
data-diff db1 table1 table2        # joindiff
data-diff db1 table1 db2 table2  -a joindiff  # also joindiff, throws error if db1 != db2
```

TODO:
- More tests for the joindiff algorithm
- Update readme docs


leoebfolsom:
I just tried out `joindiff` locally using postgres, and it worked!

fwiw, I'm not clear on why I would ever write `data-diff db1 table1 db2 table2  -a joindiff` when I could equally write `data-diff db1 table1 table2`. If there's a specific use case for the more verbose command, an example or explanation of that use case in the readme would help.

I'm happy to help workshop that or discuss further if my comment is unclear.

erezsh:
@leoebfolsom  Thanks for trying it out!

> fwiw, I'm not clear on why I would ever write data-diff db1 table1 db2 table2 -a joindiff 

Usually you wouldn't, I agree. But the idea is that in the future, you could do `-a auto` and data-diff will auto-choose the best algorithm, based on whether db1=db2.

So, it's more of a side-effect of the design, and not the recommended way to do it.

leoebfolsom:
> @leoebfolsom Thanks for trying it out!
> 
> > fwiw, I'm not clear on why I would ever write data-diff db1 table1 db2 table2 -a joindiff
> 
> Usually you wouldn't, I agree. But the idea is that in the future, you could do `-a auto` and data-diff will auto-choose the best algorithm, based on whether db1=db2.
> 
> So, it's more of a side-effect of the design, and not the recommended way to do it.

Thanks @erezsh ! I'll dig into the `-a` option and give it a test drive.

erezsh:
@leoebfolsom Sure! Just keep in mind this is still work in progress.

The latest version is in the [joindiff](https://github.com/datafold/data-diff/tree/joindiff) branch. (no PR for it yet)

It can perform the outerjoin in multiple threads, by splitting the table, so it should perform much better on postgres now. (Especially if you set `validate_unique_key = False` for `JoinDiffer`, because this validation is the only full table scan we perform.)

erezsh:
> Ideally, I could do something like data-diff db1 schema.table1 schema.table2

@williebsweet  I believe that should work. If it doesn't, open an issue with the details, and I'll make it happen.

williebsweet:
> @williebsweet I believe that should work. If it doesn't, open an issue with the details, and I'll make it happen.

@erezsh You're right - can confirm that it works. What threw me off was that the URI requires a schema, but aren't limited to that schema and can specify with the "." syntax.

leoebfolsom:
@erezsh thanks for the tip about the `joindiff` branch! 🙌

erezsh:
@williebsweet Yes, we require a default schema, but table paths can override it. I can look into making it optional, but I think we need it for creating temporary tables.

pik94 (approved):

erezsh:
Continued in #242

PR #237, opened 2022-09-16
Added a "we are hiring" paragraph to the README

erezsh:
View the new README here: https://github.com/datafold/data-diff/tree/we-are-hiring#readme


glebmezh:
Made some tweaks – good to go!

erezsh:
I can't believe there's an emoji for dollars with wings. Truly we live in the future.

PR #241, opened 2022-09-27
README updates for coalesce and pre release

leoebfolsom:
Per my discussion with @williebsweet, I'm putting together some ideas for making the README (and eventually the datafold docs) more accessible and digestible for a less technical first-time user.


erezsh:
Overall I'm okay with these changes, but I think it was good to have the benchmark graph prominent, since data-diff's main feature is its speed. I don't think it's a good idea to push it down to the technical explanation.

Ideally, we should re-run the benchmarks to include the in-db diff (joindiff), and we can re-write the explanation for the benchmark so it's a little less technical.

leoebfolsom:
Thanks for the feedback @erezsh, I appreciate it!

Just to be clear, I don't think we should merge this yet--it was meant as a work in progress and first draft so I could collect feedback and make updates. 

I'd love to schedule some time to meet and chat about this all next week, when I start full-time as a Solutions Engineer.

I agree with you that we should re-run benchmarks with joindiff. Is that something you'd be able to do? Even if it's not presentation-ready, I'd be interested in the results.

In addition to the `count(*)` comparison that's currently in the benchmark chart, it would be good to see how joindiff compares to cross-db-diff when applied to two tables in the same database. As we know, people have been using cross-db-diff to do within-db-diff (that was my original use case when I first tried out data-diff), and we want to highlight to existing users that we released a new version for in-db-diff, which is faster and designed for that use case.

When I start FT next week, I'll be more than happy to help where I can to develop these benchmarks and messaging.

leoebfolsom:
Thanks @erezsh ! Marked it as draft. 🙌

leoebfolsom:
Just a reminder that this is very much in flight, so if you take a look, please its WIP 🚧 👷  status in mind! 😇

dlawin (approved):

diveart (approved):
ok

PR #263, opened 2022-10-21
Update documentation link

williebsweet:
Switch from the current readthedocs documentation to the Datafold hosted documentation, which is also [open-source](https://github.com/datafold/datafold-docs/tree/main/docs/os_diff).

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ data-diff is a **free, open-source tool** that enables data professionals to det
 
 _Are you a developer with a deep understanding of databases and solid Python knowledge? [We're hiring!](https://www.datafold.com/careers)_
 
-[**Documentation on readthedocs.io**](https://data-diff.readthedocs.io/en/latest/)
+[**Check out our documentation!**](https://docs.datafold.com/os_diff/about)
 
 ## Use cases
 


erezsh:
Page not found at root path: https://docs.datafold.com/os_diff/

Guides still mentions dbt: https://docs.datafold.com/guides/ci_guides_overview

imo The link to Join Us On Github should be just the repo https://github.com/datafold/data-diff, and not link the the README

erezsh:
Also, is it possible to use git submodule, or some other mechanism, so that the docs can stay on the same repo as the code? Having to do a separate docs PR, in a different repo from the code PR, will make keeping them in sync much harder.

leoebfolsom:
@williebsweet I've added the Python API Reference and New Database Driver instructions to the datafold docs: https://github.com/datafold/datafold-docs/pull/32

erezsh:
@leoebfolsom There are some problems with this version of the Python API

- There is no side menu for quickly jumping around
- there are no links. for example, in "See Also", clicking object names doesn't send me to them.
- Most importantly, this is not auto-generated. How will I keep it in sync with the docstrings?

williebsweet:
Resolved the above by just linking out to the readthedocs for the Python API.

@erezsh should we remove the rest of the documentation from readthedocs so that we don't have to keep both up to date?

erezsh (approved):

PR #273, opened 2022-11-02
Fix databricks

pik94:


data_diff/databases/databricks.py
```diff
@@ -84,14 +80,21 @@ def query_table_schema(self, path: DbPath) -> Dict[str, tuple]:
...
+
         schema, table = self._normalize_table_path(path)
-        with self._conn.cursor() as cursor:
-            cursor.columns(catalog_name=self.catalog, schema_name=schema, table_name=table)
-            rows = cursor.fetchall()
+        with conn.cursor() as cursor:
+            cursor.columns(catalog_name=self._args['catalog'], schema_name=schema, table_name=table)
+            try:
+                rows = cursor.fetchall()
+            except:
```
erezsh, line 90: Always catch a specific exception.

data_diff/databases/databricks.py
```diff
@@ -84,14 +80,21 @@ def query_table_schema(self, path: DbPath) -> Dict[str, tuple]:
         # https://docs.databricks.com/spark/latest/spark-sql/language-manual/information-schema/columns.html
         # So, to obtain information about schema, we should use another approach.
 
+        conn = self.create_connection()
```
erezsh, line 83: Why create a separate connection for querying the schema?

erezsh (changes requested):
Needs 1 change and 1 explanation.

I assume you ran all the tests with at least mysql+postgresql+databricks?

pik94:
> Needs 1 change and 1 explanation.
> 
> I assume you ran all the tests with at least mysql+postgresql+databricks?

@erezsh No, I did not. Don't they run on our CI/CD?

erezsh:
@pik94 They do not. We don't have a proper way to set it up for tests.

data_diff/databases/databricks.py
pik94: It is because of our threaded database architecture. For that cases we execute sql code via query workers. However, for databricks information about schema is downloaded not via sql queries but using specific python databricks client, and to obtain the schema, we should already have a connection in this place.

pik94: I can write a method to call in in query workers but i think it is easier to just create a connection here because workers just wraps these connections themselves.

pik94:
> @pik94 They do not. We don't have a proper way to set it up for tests.

@erezsh , ok, i will run them locally then. For now move a PR to draft then.

erezsh:
Sorry, I caused a small conflict. It should be easy to resolve, but let me know if you need help with it.

pik94:
> Sorry, I caused a small conflict. It should be easy to resolve, but let me know if you need help with it.

@erezsh np, resolved it.

erezsh:
Thanks. Please rebase and reduce commits before removing the 'draft' flag.

pik94:
> Thanks. Please rebase and reduce commits before removing the 'draft' flag.

@erezsh done. Also ran unit-tests for Databricks vs Databricks, Databricks vs PostgreSQL, Databricks vs MySQL, they were passed after some fixes.

data_diff/databases/databricks.py
```diff
@@ -61,54 +62,57 @@ def normalize_timestamp(self, value: str, coltype: TemporalType) -> str:
         return f"date_format({value}, 'yyyy-MM-dd HH:mm:ss.{precision_format}')"
 
     def normalize_number(self, value: str, coltype: NumericType) -> str:
-        return self.to_string(f"cast({value} as decimal(38, {coltype.precision}))")
+        value = f"cast({value} as decimal(38, {coltype.precision}))"
+        if coltype.precision > 0:
+            value = f"format_number({value}, {coltype.precision})"
+        return f"replace({self.to_string(value)}, ',', '')"
 
     def _convert_db_precision_to_digits(self, p: int) -> int:
```
erezsh, line 70: If you're not using this method, you should remove it.

erezsh (changes requested):
Just one small thing.

Also don't forget to run `black -l 120 data_diff/databases/databricks.py`

data_diff/databases/databricks.py
erezsh: (`_convert_db_precision_to_digits`)

pik94: It is used

pik94:
> Just one small thing.
> 
> Also don't forget to run `black -l 120 data_diff/databases/databricks.py`

Done. :)

erezsh (approved):

PR #283, opened 2022-11-09
Fix so that `--min-age` and `--max-age` computed relative to database session time

nenb:
This PR is related to a recently opened Issue (#284 ). The PR is currently in draft format, awaiting further discussion on the Issue tracker.

The PR currently attempts to minimally extend some of the original data structures to retrieve a timestamp for a database session. It also re-orders the the sequence of logic slightly in `__main__.py` so that `--min-age` and `--max-age` can be computed relative to the database session timestamps.

It's not clear to me how to handle the case where both databases have different timezones, or whether this is really worth considering. At the moment, I just log and return.

### DIFF
--- a/data_diff/__main__.py
+++ b/data_diff/__main__.py
@@ -5,6 +5,7 @@ import json
 import logging
 from itertools import islice
 from typing import Optional
+from datetime import datetime, timedelta
 
 import rich
 import click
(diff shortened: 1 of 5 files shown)


nenb:
Closing following discussion in #284

PR #287, opened 2022-11-11
Materialize: rename and reorder columns

erezsh:


leoebfolsom:
@erezsh does this also address not only the ordering, but also the point that I raised in Slack that

> column names alternately use `a/b` and `1/2` to represent the left and right tables

Will they all now materialize as `a/b` instead of `1/2`?

erezsh:
@leoebfolsom Let me know if it behaves as you'd expect, so I can merge it in.

leoebfolsom:
This works as expected--thank you! @erezsh

leoebfolsom (approved):

leoebfolsom:
@erezsh please merge when you're ready.

PR #293, opened 2022-11-15
Update typing of TableSegment().count()

MattDelac:
On my side when dst is a TableSegment object, dst.count() returns an int. Using `data-diff==0.2.8`


erezsh (approved):

erezsh:
Thanks!

PR #294, opened 2022-11-15
Black -l 120 applied to the whole codebase

jardayn:


erezsh:
Thanks, but I'll do it myself.

I'm in the middle of a big refactor, so it's inconvenient right now.

PR #298, opened 2022-11-16
Rudderstack poc

kylemcnair:
Switch tracking from Freshpaint to Rudderstack.

### DIFF
--- a/data_diff/tracking.py
+++ b/data_diff/tracking.py
@@ -16,7 +16,7 @@ import toml
 TRACK_URL = "https://hosted.rudderlabs.com/v1/track"
 START_EVENT = "os_diff_run_start"
 END_EVENT = "os_diff_run_end"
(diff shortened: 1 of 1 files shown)


kylemcnair:
Events successfully showing in Rudderstack
Screen Shot 2022-11-16 at 3 10 15 PM

kylemcnair:
I was unsure about exposing the write key and auth header in the repo. We previously exposed the Freshpaint token, but it doesn't seem like best practice.

erezsh:
@kylemcnair If you can think of a different approach, I'm all ears.

kylemcnair:
> @kylemcnair If you can think of a different approach, I'm all ears.

@erezsh I don't have something in mind, but wanted to raise it. I talked to Dan about it, and don't think it's a blocking concern. I'll keep an eye out for alternatives and can always circle back in a subsequent PR.

erezsh (approved):

PR #300, opened 2022-11-17
extract methods for stats

dlawin:
I have a use-case where it's nice to have similar output to cli `--stats` while using the data_diff Python package. To accomplish this I did the following:

I extracted the print logic in __main__ to a method under DiffResultWrapper
Added flags/logic to diff_tables to use print method


tests/test_api.py
```diff
@@ -72,3 +74,28 @@ def test_api(self):
 
         t1.database.close()
         t2.database.close()
+
+    @unittest.mock.patch("sys.stdout", new_callable=io.StringIO)
+    def test_api_print(self, mock_stdout):
+        expected = "5 rows in table A\n4 rows in table B\n1 rows exclusive to table A (not present in B)\n0 rows exclusive to table B (not present in A)\n0 rows updated\n4 rows unchanged\n20.00% difference score\n\nExtra-Info:\n  rows_downloaded = 5\n"
+        t1 = connect_to_table(TEST_MYSQL_CONN_STRING, "test_api")
```
dlawin, line 81: am I missing a step with these? I assumed the setup methods would apply

dlawin: Screen Shot 2022-11-17 at 3 08 57 PM

dlawin:
> Hi, I'm okay with extracting print_stats(), but yeah, this is too much code duplication. I'm sure it can be done with less code.

How about a shared _setup_table_diff() method in __init__.py?

erezsh:
If you look at `diff_table.py`, you'll see `TableDiffer.diff_tables()` returns `DiffResultWrapper`. Maybe you should move the print logic as a method if it?

dlawin:
> If you look at `diff_table.py`, you'll see `TableDiffer.diff_tables()` returns `DiffResultWrapper`. Maybe you should move the print logic as a method if it?
> 
> i.e. `diff_wrapper.print_stats(json=json_result, ...)`

updated

erezsh:
Don't you think it's better if `diff_tables()` got a `print_stats` boolean argument?

dlawin:
> Don't you think it's better if `diff_tables()` got a `print_stats` boolean argument?

I do think that would be better, we'd have to change the current return type though since the print method iterates through the diffs

dlawin:
I think I fixed the table name collisions:
```
    unittest-parallel -v -s . -p test_api.py
Running 3 test suites (3 total tests) across 3 processes

test_api_print (test_api.TestApi) ...
test_api (test_api.TestApi) ...
test_api_print_json (test_api.TestApi) ...
test_api_print_json (test_api.TestApi) ... ok
test_api_print (test_api.TestApi) ... ok
test_api (test_api.TestApi) ... ok

Ran 3 tests in 0.399s

OK
```

data_diff/__init__.py
```diff
@@ -138,5 +144,15 @@ def diff_tables(
         )
     else:
         raise ValueError(f"Unknown algorithm: {algorithm}")
+    
+    # no key_columns provided, use table segment key_columns
+    # filter to unique values
+    if key_columns is None:
+        key_columns = list(set(list(segments[0].key_columns + segments[1].key_columns)))
```
erezsh, line 151: This is a bug.

We don't actually care about the key_columns, print_stats only wants to know *how many* key columns we have (i.e. 1 or more), and that has to be the same for both segments.

data_diff/__init__.py
```diff
@@ -85,6 +89,8 @@ def diff_tables(
         max_threadpool_size (int): Maximum size of each threadpool. ``None`` means auto.
                                    Only relevant when `threaded` is ``True``.
                                    There may be many pools, so number of actual threads can be a lot higher.
+        print_stats (bool): Enable/disable printing stats to stdout
```
erezsh, line 92: Mention that it consumes the entire iterable before returning. i.e. won't return immediately anymore.

data_diff/__init__.py
```diff
@@ -65,6 +65,10 @@ def diff_tables(
     # Maximum size of each threadpool. None = auto. Only relevant when threaded is True.
     # There may be many pools, so number of actual threads can be a lot higher.
     max_threadpool_size: Optional[int] = 1,
+    # Enable printing stats to stdout
+    print_stats: bool = False,
+    # Print stats using json, print_stats should also be True
+    print_json: bool = False,
```
erezsh, line 71: Do we really need to print the JSON? We can just return a JSON, and it's very easy to print.

erezsh (changes requested):
Looking better!

data_diff/__init__.py
dlawin: https://github.com/datafold/data-diff/pull/300/commits/2f709c52b8b2866730e92124b1a4fe4d413a16a8

data_diff/diff_tables.py
```diff
@@ -82,10 +84,62 @@ def _run_in_background(self, *funcs):
...
     info_tree: InfoTree
+    stats: dict
 
     def __iter__(self):
         return iter(self.diff)
 
+    def print_stats(self, json_output: bool):
+
+        diff_by_key = {}
+        for sign, values in self.info_tree.info.diff:
```
dlawin, line 95: @erezsh I think I'm close but when running from CLI this `self.info_tree.info.diff` is None. I think it could be a race condition?

erezsh: You should be using `self.diff`. Info-tree only collects diff in the leaf nodes, not the branches/roots

dlawin: got it ty

dlawin:
@erezsh I'm going to rework this to return json or a string. But close to complete

erezsh:
Why not make two methods, one for text and one for JSON? They can use a shared method of computing the particulars.

dlawin:
> Why not make two methods, one for text and one for JSON? They can use a shared method of computing the particulars.

I guess I'm just not seeing the downside. It would make unit testing easier, but I the repo only contains functional tests currently (from what I've seen)

erezsh:
The downside is that it's less flexible. If in the future we'll want to add the option to produce a YAML (just for example), we'll have to turn the 2-tuple into a 3-tuple, which is more likely to break someone's code.

If it's done through separate functions, we can add as many as we like. The extra computation doesn't really matter, and if it does we can easily cache it.

dlawin:
> The downside is that it's less flexible. If in the future we'll want to add the option to produce a YAML (just for example), we'll have to turn the 2-tuple into a 3-tuple, which is more likely to break someone's code.
> 
> If it's done through separate functions, we can add as many as we like. The extra computation doesn't really matter, and if it does we can easily cache it.

Good point, I'll extract methods

erezsh:
Don't forget to run black on the files you changed.

erezsh:
@dlawin In case you're stuck, the problem is with `dict[str, int]`. It's newer syntax, and isn't supported in Python 3.8 or below.

You just need to change it to `Dict[str, int]`

PR #304, opened 2022-11-20
handle a region param in snowflake connector

pik94:


erezsh:
Do you think it always happens with `us-west` specifically, or is it because the instance you're trying to connect to is in us-west?

pik94:
> Do you think it always happens with `us-west` specifically, or is it because the instance you're trying to connect to is in us-west?

Tbh, do not know whether it depends on the cluster location or not. But in our in-app product we process this case the same way because it leads to errors when connecting if `us-west` is set as a region.

erezsh:
According to this page, the `region` parameter is deprecated: https://docs.snowflake.com/en/user-guide/python-connector-api.html

Anyway, it sounds like if the region is "us-west", then you just shouldn't specify it? It doesn't look like something we should add to data-diff.

pik94:
> Anyway, it sounds like if the region is "us-west", then you just shouldn't specify it?

That's right.

> It doesn't look like something we should add to data-diff.

Ok, you may close the PR then.

PR #310, opened 2022-11-22
Fix readme link

dlawin:
Noticed this wasn't pointing to the correct repo


leoebfolsom:
Good catch @dlawin 😭

leoebfolsom (approved):
@erezsh please note this quick fix

leoebfolsom:
I guess you don't need a `git blame` for something like this.

erezsh (approved):

PR #330, opened 2022-12-04
fix link to docs

leoebfolsom:


erezsh:
I think the link is more correct the way it is now.

erezsh:
What I mean, we should change it to https://docs.datafold.com/os_diff/how_to_use/how_to_use_with_command_line

Although I consider it a big flaw that https://docs.datafold.com/os_diff/how_to_use/ is an invalid URL

leoebfolsom:
@erezsh please change it to https://docs.datafold.com/os_diff/how_to_use/how_to_use_with_command_line if you think that's better and then merge, and then we can discuss on Monday. We cannot wait all weekend to change a broken link in the readme, and I'm not available to debate right now. Thank you.

leoebfolsom:
@erezsh since I didn't hear from you, I updated the link based on your request.

erezsh:
Thanks!

PR #331, opened 2022-12-05
support 3 part identifier for snowflake

dlawin:
Snowflake: Allows specifying database as well when t1, t2 share a connection. Example:
```python
t1 = connect_to_table(CONN_INFO, ("DEV", "DEV_SCHEMA", "TABLE"), "ID")
t2 = connect_to_table(CONN_INFO, ("PROD", "PROD_SCHEMA", "TABLE"), "ID")
diff = diff_tables(t1, t2, algorithm=Algorithm.JOINDIFF)
```
or
```python
t1 = connect_to_table(CONN_INFO, "DEV.DEV_SCHEMA.TABLE", "ID")
t2 = connect_to_table(CONN_INFO, "PROD.PROD_SCHEMA.TABLE", "ID")
diff = diff_tables(t1, t2, algorithm=Algorithm.JOINDIFF)
```


data_diff/sqeleton/databases/snowflake.py
```diff
@@ -142,8 +142,27 @@ def _query(self, sql_code: Union[str, ThreadLocalInterpreter]):
...
+
+        return (
+            "SELECT column_name, data_type, datetime_precision, numeric_precision, numeric_scale "
+            f"FROM {database + '.' if database else ''}information_schema.columns "
+            f"WHERE table_name = '{name}' AND table_schema = '{schema}'"
+        )
+
+    def _normalize_table_path(self, path: DbPath) -> DbPath:
+        if len(path) == 1:
+            if self.default_schema:
```
erezsh, line 156: What if default_schema is None?

data_diff/sqeleton/databases/snowflake.py
```diff
@@ -142,8 +142,27 @@ def _query(self, sql_code: Union[str, ThreadLocalInterpreter]):
...
 
     def select_table_schema(self, path: DbPath) -> str:
-        schema, name = self._normalize_table_path(path)
-        return super().select_table_schema((schema, name))
+        """Provide SQL for selecting the table schema as (name, type, date_prec, num_prec)"""
+        database, schema, name = self._normalize_table_path(path)
+
+        return (
+            "SELECT column_name, data_type, datetime_precision, numeric_precision, numeric_scale "
+            f"FROM {database + '.' if database else ''}information_schema.columns "
```
erezsh, line 150: Please extract the table path to `columns_schema_path` (or something like that) and use `'.'.join(columns_schema_path)`. I think it would be a bit cleaner.

I hope that was clear.

erezsh (changes requested):

data_diff/sqeleton/databases/snowflake.py
dlawin: this is the logic from base.py https://github.com/datafold/data-diff/blob/6f5417a179120dbf7d4eb96055eaf8670e0e5979/data_diff/sqeleton/databases/base.py#L415-L419

When schema is None in the constructor, an exception is thrown here https://github.com/datafold/data-diff/blob/6f5417a179120dbf7d4eb96055eaf8670e0e5979/data_diff/sqeleton/databases/snowflake.py#L114
```
File "/Users/dan/repos/data-diff/data_diff/sqeleton/databases/snowflake.py", line 114, in __init__
    assert '"' not in schema, "Schema name should not contain quotes!"
TypeError: argument of type 'NoneType' is not iterable
```

That is an inaccurate exception, and is an example of where unit tests would be useful vs. the current tests, but it's also not related to the change.

erezsh: You make a good point, but the current implementation in `base.py` is actually a bug, which our drivers just happen to protect against.

You don't have to fix it in this PR, but you also shouldn't copy that bug into a new function.

If you expect `_normalize_table_path()` to return a 3-tuple, then it should always do so.

erezsh: FWIW, I'm now fixing the base implementation: https://github.com/datafold/data-diff/pull/333

dlawin: updated

data_diff/sqeleton/databases/snowflake.py
dlawin: Is this what you were thinking?
[2cea813](https://github.com/datafold/data-diff/pull/331/commits/2cea813e5e801cba2d965e6b25feb5556953e508)

PR #337, opened 2022-12-06
[CI-149] bigquery github action

AnthonyWC:
- Added IAM service account creds & project id in secrets
- added a sample file

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,11 +42,29 @@ jobs:
       - name: Install package
         run: "poetry install"
 
+      # BigQuery start
+      - id: 'auth'
+        uses: 'google-github-actions/auth@v1'
+        with:
+          credentials_json: '${{ secrets.GOOGLE_CREDENTIALS }}'
+
+      - name: 'Set up BigQuery Cloud SDK'
+        uses: 'google-github-actions/setup-gcloud@v1'
+
+    #   - name: 'Use gcloud CLI'
+    #     run: "gcloud config configurations list"
+
+      - name: "Install BigQuery for Python"
+        run: poetry add google-cloud-bigquery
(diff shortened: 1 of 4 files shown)


will:
> [CI-149 Data-diff CI](https://linear.app/datafold/issue/CI-149/data-diff-ci)
> **Why and what**
> 
> We currently don't have test coverage for the following data sources:
> 
> * BigQuery
> * Databricks
> * Oracle
> * Redshift
> 
> To resolve this, we need the following:
> 
> 1. We need access to the above DBs from GitHub CI (like we do for Snowflake, Presto, etc.)
> 2. The above DBs also need to be available for three repos: [data-diff](https://github.com/datafold/data-diff), [sqeleton](https://github.com/datafold/sqeleton), and [data-pii](https://github.com/datafold/data-pii)
> 3. Create a separate cloud solution for running the full test suite, which we can run once a week and before releases. (Context: The solution should be able to run long-running processes (approx 1-2 hours per run) on all of these databases. The current, Github CI solution drops processes after around 30 mins, at least the way it's set up now.)
> 
> **Supporting material** (links, related product/UX feedback, screenshots, documents, logs, stack traces)
> 
> Corresponding slack threads in [os-data-diff](https://datafold.slack.com/archives/C03C1JT59SA/p1668607976342889) and [team-infra](https://datafold.slack.com/archives/C028UJV3FT5/p1668782548695219)
> 
> **Requesters:** @will & @erez

For the record I did not request this. I would never

AnthonyWC:
> For the record I did not request this. I would never

Sorry this was due to linear's bot integration which was using username and username collision with Github

erezsh (approved):

erezsh:
Thanks @AnthonyWC !

PR #341, opened 2022-12-09
Add redshift external tables

danieldiamond:
## Description

Existing `data-diff` redshift capabilities are unable to read external tables. These are found in the `svv_external_columns` table. 

Original thought was to add a database flag `-e --external-table` but I don't believe flags can be specific to one database connection and not the other. A simpler approach is to simply query the external tables in redshift if none was found in the redshift one.

* Adding `select_external_table_schema` in redshift sqeleton
* Add condition to query redshift external tables if no results were found in the redshift database.

### DIFF
--- a/data_diff/sqeleton/databases/base.py
+++ b/data_diff/sqeleton/databases/base.py
@@ -339,8 +339,6 @@ class Database(AbstractDatabase):
 
     def query_table_schema(self, path: DbPath) -> Dict[str, tuple]:
         rows = self.query(self.select_table_schema(path), list)
-        if not rows and self.name == 'Redshift':
-            rows = self.query(self.select_external_table_schema(path), list)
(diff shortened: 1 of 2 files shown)


erezsh:
@danieldiamond  Made a small change so the code is more modular.

https://github.com/datafold/data-diff/pull/345

Please check that it works for you, and I'll merge it in.

If you can write a test for it, that would be even better!  (maybe [here](https://github.com/datafold/data-diff/blob/master/tests/test_api.py))

danieldiamond:
@erezsh was running into issues with the `query_table_schema` function, updated here

erezsh:
(don't worry about the CI error, it's a problem on our end)

erezsh:
Thank you for contributing!

PR #342, opened 2022-12-09
Add 'where' argument to Python API

leoebfolsom:
The `where` argument was not available in the Python API. This fixes that error of omission to create behavior that's consistent with a) the CLI behavior, and b) the Python API documentation.


leoebfolsom:
@erezsh I think I've successfully rebased and flattened into one commit. Please review, thanks!

PR #345, opened 2022-12-09
Adjustments to PR #341

erezsh:


danieldiamond:
@erezsh have you run this locally? running into `[09:07:05] ERROR - 'tuple' object has no attribute 'lower'` error and can't seem to find source of error

danieldiamond:
@erezsh unfortunately there's no stacktrace provided
`[09:07:05] ERROR - 'tuple' object has no attribute 'lower'`
this is the only information provided, possibly due to the `try except` that you've included in the PR

PR #346, opened 2022-12-12
add usage type tracking

kylemcnair:
**Summary**
This PR adds tracking to differentiate diffs triggered by CLI vs Python API

**Detail**
* adds `set_usage_mode(str)` to `tracking.py`
* sets usage mode to "CLI" and "Python API" in `__main__` and `__init__` respectively
* passes `usage mode` as event properties

I called it "usage mode", but am very very open to alternative naming conventions if something better communicates this property.

**Testing**
Example Python API diff:
Screen Shot 2022-12-12 at 9 03 50 AM

Example CLI diff
Screen Shot 2022-12-12 at 9 25 44 AM

**Formatting**
```
data-diff [track_cli_vs_python●] black data_diff/tracking.py                                            
All done! ✨ 🍰 ✨
1 file left unchanged.
```

### DIFF
--- a/data_diff/__main__.py
+++ b/data_diff/__main__.py
@@ -20,7 +20,7 @@ from .sqeleton.queries.api import current_timestamp
 from .databases import connect
 from .parse_time import parse_time_before, UNITS_STR, ParseError
 from .config import apply_config_from_file
-from .tracking import disable_tracking
+from .tracking import disable_tracking, set_entrypoint_name
 from .version import __version__
 
 
@@ -32,6 +32,7 @@ COLOR_SCHEME = {
     "-": "red",
 }
 
+set_entrypoint_name("CLI")
 
 def _remove_passwords_in_dict(d: dict):
     for k, v in d.items():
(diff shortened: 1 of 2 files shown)


erezsh:
We can call it "entrypoint name". 

Btw, just set the default value to "api". We only need change the value when running from main.

kylemcnair:
> We can call it "entrypoint name".
> 
> Btw, just set the default value to "api". We only need change the value when running from main.

That makes sense, updated 👍

erezsh:
Great. Any chance you can flatten them into 1 commit? (using rebase)

Also, a tiny detail, better to change the `str` param to something else, since `str` is a built in type, and it might be a little confusing. You can rename it to `name` or even `s`.

kylemcnair:
> Great. Any chance you can flatten them into 1 commit? (using rebase)
> 
> Also, a tiny detail, better to change the `str` param to something else, since `str` is a built in type, and it might be a little confusing. You can rename it to `name` or even `s`.

Done!

erezsh (approved):

PR #348, opened 2022-12-12
allow joindiff across bigquery projects

segv:


segv:
> Why not just follow the simple way we implemented it for snowflake, where you use the same database connection, **but then specify a different project in the table path?**

My brain just didn't make the connection that this was what you were talking about. PR updated.

erezsh:
Looking better! I believe the change in `info_tree.py` isn't necessary anymore, right?

erezsh:
(don't worry about the CI error, it's a problem on our end)

segv:
I reverted the `rowcounts: Dict[int, int] = {1: 0, 2: 0}` change.

erezsh:
Thank you for contributing!

PR #355, opened 2022-12-21
Test extending alphanumerics

erezsh:


RoderickJDunn:
Hello! I'd be interested in understanding the context/plans for this PR. We have several tables containing a period in their PK string column, so these currently can't be diffed.

erezsh:
The plans are to ensure that this change doesn't break or significantly reduces performance for regular alphanumeric columns.

Once we verify this, ideally through our own tests, but I would also be satisfied by accounts of several users who tried it and reported back, I'll feel comfortable merging it in.

Issue #278, opened 2022-11-04
Support environment variables in TOML files


erezsh:
Probably the best approach is to implement it by doing string interpolation on all string values in the returned dict.

PR #356, opened 2022-12-24
Add env variables support in .toml config

vvkh:
Add env variables support in `.toml` config (https://github.com/datafold/data-diff/issues/278).

Obvious syntax options were:
- `{ENV_VAR}` like Python f-strings or `.format()`.
- `$ENV_VAR` like in shell scripts.
- `${ENV_VAR}` is also widely used. For example, in [docker-compose](https://docs.docker.com/compose/environment-variables/).

I suggest using `${ENV_VAR}`, mainly because: 
- The f-strings-like approach is something I find very Python-specific.
- In contrast to `$ENV_VAR`, we can probably get away without escaping special characters with this approach.

However, I don't mind switching format if, for some reason, `${ENV_VAR}` doesn't work for us.

### DIFF
--- a/data_diff/config.py
+++ b/data_diff/config.py
@@ -1,3 +1,5 @@
+import re
+import os
 from typing import Any, Dict
 import toml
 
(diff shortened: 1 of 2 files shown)


pik94 (approved):

leoebfolsom (approved):
My review is not technical, but I fully support this change and the approach described in the PR description. 🙌 🥳

PR #364, opened 2023-01-05
add dbt integration

dlawin:
Adds options to datadiff: `data-diff --dbt` and `data-diff --dbt --cloud`

With some suboptions:
 --dbt-profiles-dir
 --dbt-project-dir

The main options parse a dbt project's artifacts and profiles.yml in order to easily run multiple diffs without specifying the typical connection strings and table names.

Currently, it will run a diff for each successful model in the project's last run.

The main configuration step is to add one or both of the following to the vars section of the dbt_project.yml:
```
vars:
  data_diff:
    prod_database: PROD_DATABASE_NAME
    prod_schema: PROD_SCHEMA_NAME
```
If a dbt project is organized at a schema level (e.g. all models write to a single schema), both are needed. If the dbt project writes to multiple schemas, just the database var is needed.

If you're using the --dbt-cloud option, that requires a Datafold api key on the env var DATAFOLD_API_KEY. Additionally a datasource_id in the vars section:
```
vars:
  data_diff:
    prod_database: PROD_DATABASE_NAME
    prod_schema: PROD_SCHEMA_NAME
    datasource_id: 123
```

Caveats:

- Limited database support (Snowflake, BigQuery so far)
- Dbt version >= 1.0.0
    - Other versions may be fine if the format of the artifacts don't change dramatically. I have not tested < 1.0.0
- I've only added unit tests for the most part. There are two integration tests that can be used to debug a "real" dbt diff session, but you need to set an env variable to enable them.
- Cloud diffs just provide a link to the diff running on Datafold (no results)
- Local diffs run sequentially, could improve performance by running in parallel


erezsh:
This change of syntax is a breaking change for every user of data-diff, not to mention our own documentation and videos. 

Why not add a `--dbt` flag instead?

dlawin:
> This change of syntax is a breaking change for every user of data-diff, not to mention our own documentation and videos.
> 
> Why not add a `--dbt` flag instead?

Mainly to support flags for the dbt functionality. I think it would be confusing to have flags that only apply when the `--dbt` flag is present. This would be odd for example:
`data-diff --dbt --cloud`

What would the behavior of `data-diff --cloud` or `data-diff -proj /path` be?

leoebfolsom:
This is very minor, but we should be consistent around saying database/table 1/2 vs database/table a/b. We should use a/b instead of 1/2.

Erez will recall that I was very annoying about this when he was naming columns for materialized tables. While I don't feel strongly about 1 vs a, we settled on a, and that is reflected in this guide: https://docs.datafold.com/guides/os_diff

🙏

erezsh:
@dlawin 

> I think it would be confusing to have flags that only apply when the --dbt flag is present

We already have this kind of behavior for hashdiff vs joindiff. It might be confusing, but I don't know of an easy way to avoid it.

Also, I believe we can allow the `data-diff dbt ...` without making any of these changes.

We already do a bit of syntax shenanigans [here](https://github.com/datafold/data-diff/blob/master/data_diff/__main__.py#L204) to allow the `db table1 table2` syntax. It shouldn't be hard to tweak it, so it tests if `database1 == "dbt"` and acts accordingly.

The last alternative is to create a new entry-point, like `data-diff-dbt ...`, which I think is a bit of an overkill, but might be a bit less confusing.

Either way, breaking the API is a definite no-no.

dlawin:
> @dlawin
> 
> > I think it would be confusing to have flags that only apply when the --dbt flag is present
> 
> We already have this kind of behavior for hashdiff vs joindiff. It might be confusing, but I don't know of an easy way to avoid it.
> 
> Also, I believe we can allow the `data-diff dbt ...` syntax without making any of these changes.
> 
> We already do a bit of syntax shenanigans [here](https://github.com/datafold/data-diff/blob/master/data_diff/__main__.py#L204) to allow the `db table1 table2` syntax. It shouldn't be hard to tweak it, so it tests if `database1 == "dbt"` and acts accordingly.
> 
> The last alternative is to create a new entry-point, like `data-diff-dbt ...`, which I think is a bit of an overkill, but might be a bit less confusing.
> 
> Either way, breaking the API is a definite no-no.

Switched to option flags, dbt specific functionality is prefixed by --dbt-

dlawin:
I did move some of the option functionality up to the main method when I thought it made sense (--debug, --no-tracking, etc.)

For now, the majority of the other options are ignored when using --dbt or --dbt-cloud, but we can add that over time

data_diff/__main__.py
```diff
@@ -200,24 +243,74 @@ def write_usage(self, prog: str, args: str = "", prefix: Optional[str] = None) -
...
+@click.option(
+    "--dbt",
+    is_flag=True,
+    help="Run a diff using your local dbt project. Expects to be run from a dbt project folder by default.",
+)
+@click.option(
+    "--dbt-cloud",
+    is_flag=True,
+    help="Run a diff using your local dbt project on Datafold cloud. Expects an api key on env var DATAFOLD_API_KEY.",
+)
```
nolar, line 255: Wouldn't it be better to run this mode as `data-diff --dbt --cloud` — since there is already the `--cloud` option with this exact meaning?

data_diff/dbt.py
```diff
@@ -0,0 +1,313 @@
...
+from dbt_artifacts_parser.parser import parse_run_results, parse_manifest
+from . import connect_to_table, diff_tables, Algorithm
+from dbt.config.renderer import ProfileRenderer
+
+CONST_RUN_RESULTS_PATH = "/target/run_results.json"
+CONST_MANIFEST_PATH = "/target/manifest.json"
+CONST_PROJECT_FILE = "/dbt_project.yml"
+CONST_PROFILES_FILE = "/profiles.yml"
+CONST_DBT_V_ABOVE = "1.0.0"
+CONST_DBT_V_BELOW = "1.4.0"
```
nolar, line 21: **nit:** There is no need to explicitly mention `CONST_…` for constants — it is already clear from the conventional name (all-caps). Just `RUN_RESULTS_PATH` would be fine.

data_diff/dbt.py
```diff
@@ -0,0 +1,313 @@
...
+        config_prod_database = datadiff_variables.get("prod_database")
+        config_prod_schema = datadiff_variables.get("prod_schema")
+        datasource_id = datadiff_variables.get("datasource_id")
+
+        if not is_cloud:
+            dbt_parser.set_connection()
+
+        assert (
+            config_prod_database is not None or config_prod_schema is not None
+        ), "Expected a value for prod_database: or prod_schema: under \nvars:\n  data_diff: "
```
nolar, line 49: It is better to use explicit `if` & `raise` for input validation or business logic checks. Assertions can be disabled if the users prefer to run Python in "optimized" mode (`python -O`), which can be the case if the library is used not via CLI but is embedded into the users' toolchain (for example, I used it for data diffs in Alembic migrations, though it was not for dbt):

```
$ python -O -c 'assert False'
(no output)
```

```
$ python -c 'assert False'
  File "<string>", line 1, in <module>
AssertionError
```

data_diff/dbt.py
```diff
@@ -0,0 +1,313 @@
...
+        dev_schema = model.schema_
+        primary_keys = dbt_parser.get_primary_keys(model)
+
+        if config_prod_database is None:
+            prod_database = dev_database
+        else:
+            prod_database = config_prod_database
+
+        if config_prod_schema is None:
+            prod_schema = dev_schema
```
nolar, line 92: Just to clarify: Somewhere above in the PR, it says that the schema should not be specified if it is multi-schema diff — even for production database. However, here it tries to use the dev schema as the prod schema even if the prod database is specified properly but with not schema. Is this intended?

data_diff/dbt.py
```diff
@@ -0,0 +1,313 @@
...
+        prod_qualified_string = ".".join(diff_vars.prod_path)
+        primary_key = diff_vars.primary_keys[0]
+
+        table1 = connect_to_table(diff_vars.connection, dev_qualified_string, primary_key)
+        table2 = connect_to_table(diff_vars.connection, prod_qualified_string, primary_key)
+
+        table1_columns = list(table1.get_schema())
+        try:
+            table2_columns = list(table2.get_schema())
+        # TODO add a unit test for scenario
```
nolar, line 118: In case it was overlooked: reminding about a TODO ;-)

data_diff/dbt.py
```diff
@@ -0,0 +1,313 @@
...
+        except Exception as e:
```
nolar, line 119: Is there a specialized exception class escalating from the database connectors? Because `Exception` can be too generic — e.g. a division by zero or type mismatches. That will be highly confusing to see an error about the table access in such cases.

data_diff/dbt.py
```diff
@@ -0,0 +1,313 @@
...
+
+        mutual_set = set(table1_columns) & set(table2_columns)
+        table1_set_diff = list(set(table1_columns) - set(table2_columns))
+        table2_set_diff = list(set(table2_columns) - set(table1_columns))
+
+        if len(table1_set_diff) > 0:
+            column_diffs_str += "Columns exclusive to table A: " + str(table1_set_diff) + "\n"
+
+        if len(table2_set_diff) > 0:
+            column_diffs_str += "Columns exclusive to table B: " + str(table2_set_diff) + "\n"
```
nolar, line 140: **nit:** Since the tool requires Python 3, it is safe to use the f-strings here for readability:

```suggestion
        if len(table1_set_diff) > 0:
            column_diffs_str += f"Columns exclusive to table A: {table1_set_diff}\n"

        if len(table2_set_diff) > 0:
            column_diffs_str += f"Columns exclusive to table B: {table2_set_diff}\n"
```

data_diff/dbt.py
```diff
@@ -0,0 +1,313 @@
...
+
+        mutual_set.discard(primary_key)
+        extra_columns = tuple(mutual_set)
+
+        diff = diff_tables(table1, table2, threaded=True, algorithm=Algorithm.JOINDIFF, extra_columns=extra_columns)
+
+        if len(list(diff)) > 0:
```
nolar, line 147: Mind that the difference can be huge (e.g. thousands or millions of rows). This place (`list(diff)`) will try to materialize them all into memory before doing a check.

Perhaps, you want to only check if it yields at least one item with `next(diff)`. If it is empty, it will raise `StopIteration`. Alternatively, `next(diff, None)` can be used to return `None` (or any other placeholder) if the diff is empty, the first item otherwise. In that case, it will get one diff item, abandon the object, and it will be garbage-collected almost instantly.

data_diff/dbt.py
```diff
@@ -0,0 +1,313 @@
...
+        api_key = os.environ.get("DATAFOLD_API_KEY")
+
+        if diff_vars.datasource_id is None:
+            raise ValueError(
+                "Datasource ID not found, include it as a dbt variable in the dbt_project.yml. \nvars:\n data_diff:\n   datasource_id: 1234"
+            )
+        elif api_key is None:
+            raise ValueError("API key not found, add it as an environment variable called DATAFOLD_API_KEY.")
+        else:
+            url = "https://app.datafold.com/api/v1/datadiffs"
```
nolar, line 180: **nit:** It would be good to parametrize the hostname and prefix too — in case the user uses a self-hosted solution. (But not in this PR, maybe separately.)

data_diff/dbt.py
```diff
@@ -0,0 +1,313 @@
...
+                    "table2": diff_vars.prod_path,
+                    "pk_columns": diff_vars.primary_keys,
+                }
+            )
+            headers = {
+                "Authorization": "Key " + api_key,
+                "Content-Type": "application/json",
+            }
+
+            response = requests.request("POST", url, headers=headers, data=payload)
```
nolar, line 196: **nit:** `requests` accepts any serializable object and converts it to JSON itself if used as `… = requests.post(url, headers=headers, json=payload)`. It would make it more readable, and perhaps testable — but this is not critical.

data_diff/dbt.py
```diff
@@ -0,0 +1,313 @@
...
+            response.raise_for_status()
+            data = response.json()
+            id = data["id"]
+            diff_url = f"https://app.datafold.com/datadiffs/{str(id)}/overview"
```
nolar, line 200: **nit:** This is the same as:

```suggestion
            diff_url = f"https://app.datafold.com/datadiffs/{id}/overview"
```

data_diff/dbt.py
```diff
@@ -0,0 +1,313 @@
...
+                + ".".join(diff_vars.prod_path)
+                + "[/] \n    Diff in progress: \n    "
+                + diff_url
+                + "\n"
+            )
+
+
+class DbtParser:
+    DEFAULT_PROFILES_DIR = str(Path.home()) + "/.dbt"
+    DEFAULT_PROJECT_DIR = str(os.getcwd())
```
nolar, line 214: **nit:** From my personal experience using CLI tools, I prefer to see `./somefile` in my logs and error messages if I implied the current directory. With `os.getcwd()`, it will be using the absolute path of the current directory, such as `/Users/nolar/src/some-project/somefile`. Not that it is a big problem, but adds a log of visual garbage.

Similarly with the home dir: `~/.dbt/somefile` instead of `/Users/nolar/.dbt/somefile`. But that requires using `os.path.expanduser(path)` in the code below, so maybe not worth changing, but worth knowing for the future.

data_diff/dbt.py
```diff
@@ -0,0 +1,313 @@
...
+        self.project_dir = project_dir_override or self.DEFAULT_PROJECT_DIR
+        self.is_cloud = is_cloud
+        self.connection = None
+        self.project_dict = None
+        self.requires_upper = False
+
+    def get_datadiff_variables(self):
+        try:
+            vars = self.project_dict.get("vars").get("data_diff")
+            assert vars is not None
```
nolar, line 227: There can be a case when `vars:` is defined, but it is not a dict/object. It will fail with an exception anyway. So, It might be a good idea to always fail with an explanatory `KeyError` instead of a cryptic `AssertionError`.

```suggestion
            vars = self.project_dict["vars"]["data_diff"]
```

data_diff/dbt.py
```diff
@@ -0,0 +1,313 @@
...
+            logging.debug(e)
+            raise (Exception("Expected values in dbt_project.yml under \nvars:\n  data_diff:"))
+        return vars
+
+    def get_models(self):
+        with open(self.project_dir + CONST_RUN_RESULTS_PATH) as run_results:
+            run_results_dict = json.load(run_results)
+            run_results_obj = parse_run_results(run_results=run_results_dict)
+
+        # TODO need test for less than scenario
```
nolar, line 238: A reminder about a presumably forgotten TODO in the code ;-)

data_diff/__main__.py
williebsweet: Agreed.

data_diff/dbt.py
dlawin: fixed

data_diff/dbt.py
dlawin: Yes this is intended. What this does is make the schema dynamic for each model rather than fixed

If the dbt project uses database environments (so config_prod_schema is None), we are assuming that the same schema name being used in dev_db and prod_db are equivalent for each model

This assumption isn't always going to be correct, especially if they use [custom schemas](https://docs.getdbt.com/docs/build/custom-schemas) but this covers most cases in the absence of a prod manifest.json

data_diff/dbt.py
dlawin: [a0d7980](https://github.com/datafold/data-diff/pull/364/commits/a0d7980dec42969c50f339c4824ac3a0f02f08e9)

data_diff/dbt.py
dlawin: that materialization is needed in order to have stats to print with `diff.get_stats_string()` and will occur if I call that method regardless

see:
https://github.com/datafold/data-diff/blob/2d624d651991826e2bf652fd8fb85e504c6dd5d8/data_diff/diff_tables.py#L104

data_diff/dbt.py
dlawin: Unfortunately that doesn't seem to be the case. There is [this](https://github.com/datafold/sqeleton/blob/ffc4afe80cdd54b9ee29d2aa429346af7cce5a25/sqeleton/databases/base.py#L391) exception, but exceptions can be raised by the various underlying connection packages before that. And they aren't being routed to an exception type that spans them

For now, maybe I can log the exception every time so the user can interpret

data_diff/dbt.py
dlawin: added a comment to address this in future

data_diff/dbt.py
dlawin: fixed

data_diff/dbt.py
dlawin: changed

data_diff/__main__.py
dlawin: altered so that the dbt functionality works like:
`data-diff --dbt` = local dbt diff
`data-diff --dbt --cloud` = cloud dbt diff

dlawin:
tested compatibility with dbt:
1.0
1.1
1.2
1.3
1.4

I didn't cover every adapter (just snowflake), but that should not affect the dbt-core artifacts being parsed.

dlawin:
Hey @williebsweet @nolar, I think I've addressed all of the feedback. Re-requested review

data_diff/__main__.py
nolar: Hm. I thought we had the `--cloud` option to run diffs in the cloud. Maybe I imagined that instead, but that mode was in the discussion for some time, so let's go that way by including the option for dbt scenario only and extending its meaning later.

data_diff/dbt.py
nolar: Ah, I see. That is a place for a potential problem then. It should be iterating over `self` until depleted, not constructing a `list()` and immediately discarding it. 

Nevertheless, this is out of the scope of this PR and should be fixed separately. Thanks for clarifying.

nolar (approved):

pyproject.toml
```diff
@@ -38,6 +38,9 @@ trino = {version="^0.314.0", optional=true}
 presto-python-client = {version="*", optional=true}
 clickhouse-driver = {version="*", optional=true}
 duckdb = {version="^0.6.0", optional=true}
+dbt-artifacts-parser = "^0.2.4"
+dbt-core = "^1.0.0"
+filelock = "^3.9.0"
```
erezsh, line 43: Why is filelock added here?

erezsh (changes requested):

pyproject.toml
dlawin: For some reason, poetry was removing this package and it was breaking my poetry install locally

data_diff/__main__.py
```diff
@@ -13,9 +13,14 @@
 from sqeleton.schema import create_schema
 from sqeleton.queries.api import current_timestamp
 
+from .dbt import dbt_diff
 from .utils import eval_name_template, remove_password_from_url, safezip, match_like
 from .diff_tables import Algorithm
-from .hashdiff_tables import HashDiffer, DEFAULT_BISECTION_THRESHOLD, DEFAULT_BISECTION_FACTOR
```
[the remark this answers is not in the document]
dlawin: this file is formatted with `black -l 120`

```
(datadiff)     black -l 120 data_diff/__main__.py
All done! ✨ 🍰 ✨
1 file left unchanged.
```

pyproject.toml
dlawin: it's a dependency for poetry itself so I'm not sure why it would be removed

data_diff/__main__.py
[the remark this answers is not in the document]
dlawin: Oh I see what you mean. Idk why some of these <120 lines got changed... reverrting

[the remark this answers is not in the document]
dlawin: fixed

pyproject.toml
dlawin:
```
(datadiff)     poetry remove filelock
Updating dependencies
Resolving dependencies... (2.5s)

Writing lock file

Package operations: 0 installs, 0 updates, 1 removal

  • Removing filelock (3.9.0)
  
.
.
(datadiff)     poetry install
ModuleNotFoundError: No module named 'filelock'
```

dlawin: No longer seeing poetry remove this package automatically. I've reverting adding it

erezsh (approved):
Good enough for me

PR #374, opened 2023-01-28
Restructure CI output

leoebfolsom:
Restructuring CI output to resemble the Datafold app diff overview.

BEFORE:
Screenshot 2023-01-30 at 9 39 20 AM

AFTER:
Screenshot 2023-01-30 at 9 10 54 AM

### DIFF
--- a/data_diff/diff_tables.py
+++ b/data_diff/diff_tables.py
@@ -174,7 +174,7 @@ class DiffResultWrapper:
         
         # Display columns with conflicts
         columns_with_conflicts = []
-        conflicts_df = pd.DataFrame(columns=['Column', 'Conflicting Rows'])
+        conflicts_df = pd.DataFrame(columns=['Column', 'Values Updated'])
         
         for i in additional_columns_to_diff:
             conflicts = (matching_primary_key_rows[i+": Prod"] != matching_primary_key_rows[i+": Dev"]) & \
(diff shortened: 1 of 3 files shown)


data_diff/diff_tables.py
```diff
@@ -100,11 +102,67 @@ def __iter__(self):
             self.result_list.append(i)
             yield i
 
+        for sign, values in self.result_list:
```
dlawin, line 105: Could this replace the behavior of get_stats_string() instead? Or maybe be an additional method like get_stats_tables_string()?

leoebfolsom: I think it can replace `get_stats_string()`. I don't think there's any drawback at all to fully replacing the existing `string_output` since it is not structured (and therefore depended on by some downstream system)--it's just a string.

I will refactor this, and I'm very open to suggestions/recommendations, but I'm going to try to settle on a good design of the output before optimizing the code.

leoebfolsom: @dlawin moved the logic into `get_stats_string` 👍  (still could be optimized from here)

williebsweet:
@leoebfolsom this seems like a duplicate of #381. Can we close this?

leoebfolsom:
@williebsweet yes it's a duplicate, closing, ty

PR #380, opened 2023-02-06
Add events for dbt feature

dlawin:
Adds new events for the `--cloud` option

Sets a new entrypoint `CLI-dbt` for `--dbt` and `--cloud`

Added additional parameters to end_event in order to differentiate `--cloud` events

Most of the information is hardcoded to 0/None/etc. on the end event as we don't currently have the related information. Since I think it's likely we will be adding polling and diff results for `--cloud` in the near future, I thought it made sense to reuse the existing event structure.

### DIFF
--- a/data_diff/tracking.py
+++ b/data_diff/tracking.py
@@ -85,12 +85,14 @@ def create_start_event_json(diff_options: Dict[str, Any]):
 def create_end_event_json(
     is_success: bool,
     runtime_seconds: float,
-    db1: str,
-    db2: str,
+    data_source_1_type: str,
+    data_source_2_type: str,
     table1_count: int,
     table2_count: int,
     diff_count: int,
     error: Optional[str],
+    diff_id: Optional[int] = None,
+    is_cloud: bool = False
 ):
     return {
         "event": "os_diff_run_end",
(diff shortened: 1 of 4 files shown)


data_diff/utils.py
```diff
@@ -72,3 +72,8 @@ def get_timestamp(_match):
         return datetime.now().isoformat("_", "seconds").replace(":", "_")
 
     return re.sub("%t", get_timestamp, name)
+
+
+def truncate_error(error: str):
```
dlawin, line 77: moved this from diff_tables.py so I can reuse it

data_diff/tracking.py
```diff
@@ -85,12 +85,14 @@ def create_start_event_json(diff_options: Dict[str, Any]):
 def create_end_event_json(
     is_success: bool,
     runtime_seconds: float,
-    db1: str,
-    db2: str,
+    data_source_1_type: str,
```
dlawin, line 88: I changed these just because I found the names confusing/misleading

kylemcnair: non-blocking: I'll need to update downstream reporting

dlawin: this just impacts the python code, the actual event is not affected

kylemcnair: oh I see, still "data_source_1_type" 👍

kylemcnair (approved):
Screen Shot 2023-02-07 at 3 55 41 PM
Events with the new properties look good in Snowflake

Adding approval from that perspective, deferring to others tagged for review of the code

erezsh:
@dlawin Looks okay to me. Did you run black?

williebsweet:
> @dlawin Looks okay to me. Did you run black?

@erezsh Dan is on PTO today. Can you pull it down and do that?

erezsh:
Done

PR #381, opened 2023-02-06
spruce up CLI output in dbt context

leoebfolsom:
My attempt at improving the CLI printout without using any additional python packages.

BEFORE:
Screenshot 2023-02-01 at 23 54 09

AFTER:
Screenshot 2023-02-01 at 23 52 48

TODO based on comments in https://github.com/dlawin/data-diff/pull/3:

- [x] @erezsh: Looks like _get_stats_dbt() should instead just be a flag to _get_stats() https://github.com/dlawin/data-diff/pull/3#issuecomment-1413855643
- [ ] @erezsh: Better to implement this using zip() / safezip() https://github.com/dlawin/data-diff/pull/3#discussion_r1094628001
- [x] @dlawin: Since we're adding a parameter here, we'll need to account for it in the other DiffStats() in this file https://github.com/dlawin/data-diff/pull/3#discussion_r1095035933
- [x] @dlawin : update `get_stats_string(self, is_dbt:bool = False):` ... add the additional functionality there and just wrap it in an `if is_dbt:` https://github.com/dlawin/data-diff/pull/3#issuecomment-1414321053
- [x] @dlawin : line length

### DIFF
--- a/data_diff/diff_tables.py
+++ b/data_diff/diff_tables.py
@@ -104,10 +104,9 @@ class DiffResultWrapper:
     def _get_stats(self, is_dbt:bool = False) -> DiffStats:
         list(self)  # Consume the iterator into result_list, if we haven't already
 
-        
         key_columns = self.info_tree.info.tables[0].key_columns
         diff_by_key = {}
-
+        extra_column_diffs = None
         if is_dbt:
             extra_column_values_store = {}
             extra_columns = self.info_tree.info.tables[0].extra_columns
@@ -137,16 +136,13 @@ class DiffResultWrapper:
         table2_count = self.info_tree.info.rowcounts[2]
         unchanged = table1_count - diff_by_sign["-"] - diff_by_sign["!"]
         diff_percent = 1 - unchanged / max(table1_count, table2_count)
-        if is_dbt:
-            return DiffStats(diff_by_sign, table1_count, table2_count, unchanged, diff_percent, extra_column_diffs)
-        else:
-            return DiffStats(diff_by_sign, table1_count, table2_count, unchanged, diff_percent, None)
+
+        return DiffStats(diff_by_sign, table1_count, table2_count, unchanged, diff_percent, extra_column_diffs)
 
     def get_stats_string(self, is_dbt:bool = False):
 
-        diff_stats = self._get_stats()
+        diff_stats = self._get_stats(is_dbt)
         if is_dbt:
-            diff_stats = self._get_stats(is_dbt=True)
             
             string_output = "\n| Rows Added\t| Rows Removed\n"
             string_output += "------------------------------------------------------------\n"
--- a/data_diff/diff_tables.py
+++ b/data_diff/diff_tables.py
@@ -166,15 +166,7 @@ class DiffResultWrapper:
     def get_stats_string_dbt(self):
 
         diff_stats = self._get_stats_dbt()
-        string_output = ""
-        string_output += f"{diff_stats.table1_count} rows in table A\n"
-        string_output += f"{diff_stats.table2_count} rows in table B\n"
-        string_output += f"{diff_stats.diff_by_sign['-']} rows exclusive to table A (not present in B)\n"
-        string_output += f"{diff_stats.diff_by_sign['+']} rows exclusive to table B (not present in A)\n"
-        string_output += f"{diff_stats.diff_by_sign['!']} rows updated\n"
-        string_output += f"{diff_stats.unchanged} rows unchanged\n"
-        string_output += f"{100*diff_stats.diff_percent:.2f}% difference score\n"
-
+        
         string_output = "\n| Rows Added\t| Rows Removed\n"
         string_output += "------------------------------------------------------------\n"
 
--- a/data_diff/diff_tables.py
+++ b/data_diff/diff_tables.py
@@ -101,7 +101,7 @@ class DiffResultWrapper:
             self.result_list.append(i)
             yield i
 
-    def _get_stats(self) -> DiffStats:
+    def _get_stats(self, is_dbt:bool = False) -> DiffStats:
         list(self)  # Consume the iterator into result_list, if we haven't already
 
         
--- a/data_diff/diff_tables.py
+++ b/data_diff/diff_tables.py
@@ -101,7 +101,7 @@ class DiffResultWrapper:
             self.result_list.append(i)
             yield i
 
-    def _get_stats(self, is_dbt:bool = False) -> DiffStats:
+    def _get_stats(self, is_dbt: bool = False) -> DiffStats:
         list(self)  # Consume the iterator into result_list, if we haven't already
 
         key_columns = self.info_tree.info.tables[0].key_columns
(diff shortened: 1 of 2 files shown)


data_diff/diff_tables.py
```diff
@@ -85,6 +85,7 @@ class DiffStats:
     table2_count: int
     unchanged: int
     diff_percent: float
+    extra_column_diffs: Dict[str, int]
```
dlawin, line 88:
```suggestion
    extra_column_diffs: Optional[Dict[str, int]]
```
Should fix the CI error

data_diff/diff_tables.py
```diff
@@ -100,17 +101,33 @@ def __iter__(self):
             self.result_list.append(i)
             yield i
 
-    def _get_stats(self) -> DiffStats:
+    def _get_stats(self, is_dbt:bool = False) -> DiffStats:
```
leoebfolsom, line 104: @dlawin can I use the same approach here--defining `is_dbt` as an input to `_get_stats`, which is called by `get_stats_string`?

dlawin: yep

leoebfolsom:
@erezsh first, please note that this is a new PR now that Dan's dbt integration is merged into master.

I've made all the changes suggested by you and Dan in the previous PR _except for_ [using zip() / unzip()](https://github.com/dlawin/data-diff/pull/3#discussion_r1094628001). I found some explainers since `zip()` is new to me, but I'm not sure how they would apply here. 

My goal is to add to keep track of conflicts for each of the `extra_columns` as I iterate through the rows of `result_list`, and store the counts in a dictionary: `{ first_name: 14, last_name: 0, street: 0, city: 2, state: 0, ... }` representing the # of conflicts per column.

I can see how it's repetitive to iterate through `for i in range(0,len(extra_columns)):` for each row and then check the value based on the index (`if extra_column_values[i] != extra_column_values_store[k][i]:`) for every column. 

However, I'm struggling to see where I could gain efficiency using `zip()`. Can you please be more specific about what you have in mind and/or provide an example?

data_diff/diff_tables.py
```diff
@@ -120,25 +137,45 @@ def _get_stats(self) -> DiffStats:
...
-        string_output += f"{diff_stats.diff_by_sign['!']} rows updated\n"
-        string_output += f"{diff_stats.unchanged} rows unchanged\n"
-        string_output += f"{100*diff_stats.diff_percent:.2f}% difference score\n"
-
-        if self.stats:
-            string_output += "\nExtra-Info:\n"
-            for k, v in sorted(self.stats.items()):
-                string_output += f"  {k} = {v}\n"
+        if is_dbt:
+            diff_stats = self._get_stats_dbt()
```
dlawin, line 149: no need for this right?

data_diff/diff_tables.py
```diff
@@ -120,25 +137,45 @@ def _get_stats(self) -> DiffStats:
         table2_count = self.info_tree.info.rowcounts[2]
         unchanged = table1_count - diff_by_sign["-"] - diff_by_sign["!"]
         diff_percent = 1 - unchanged / max(table1_count, table2_count)
+        if is_dbt:
+            return DiffStats(diff_by_sign, table1_count, table2_count, unchanged, diff_percent, extra_column_diffs)
+        else:
+            return DiffStats(diff_by_sign, table1_count, table2_count, unchanged, diff_percent, None)
```
dlawin, line 143: if you add `extra_column_diffs = None` at the top, this can be simplified
```suggestion
        return DiffStats(diff_by_sign, table1_count, table2_count, unchanged, diff_percent, extra_column_diffs)
```

data_diff/diff_tables.py
```diff
@@ -120,25 +137,45 @@ def _get_stats(self) -> DiffStats:
...
 
-        return DiffStats(diff_by_sign, table1_count, table2_count, unchanged, diff_percent)
-
-    def get_stats_string(self):
+    def get_stats_string(self, is_dbt:bool = False):
 
         diff_stats = self._get_stats()
```
dlawin, line 147: `diff_stats = self._get_stats()`
this should move to the else so we don't run _get_stats() twice when is_dbt == True

data_diff/diff_tables.py
leoebfolsom: @dlawin I think I actually need to change it to `diff_stats = self._get_stats(is_dbt)`, right? And then remove the redundant declaration of `diff_stats` like you suggested? See latest version.

data_diff/diff_tables.py
leoebfolsom: Does my latest update accomplish the same thing? Passing the `is_dbt` variable to `_get_stats`? https://github.com/datafold/data-diff/pull/381/files#diff-b8c2ad85fce41be1450cde985be83f28803c095ba5d3f0235fc916ed8f965e80R144

data_diff/diff_tables.py
dlawin: oh yeah true

data_diff/diff_tables.py
```diff
@@ -100,17 +101,32 @@ def __iter__(self):
             self.result_list.append(i)
             yield i
 
-    def _get_stats(self) -> DiffStats:
+    def _get_stats(self, is_dbt:bool = False) -> DiffStats:
```
dlawin, line 104: Not blocking, but the logic in this method is getting very dense with these added is_dbt sections. It's pretty hard for me to follow what's going on.

With for loops nested in for loops, I wonder how this scales as diff sizes grow.

Since this is limited to the beta `--dbt` feature, I think as long as we test that diff stats are returning in \~reasonable\~ time-frames, we could tune + refactor later. This does not have test coverage currently so we should test manually with non-trivial diffs. I'd like to avoid `--dbt` hanging for any testers with medium-large diffs

data_diff/diff_tables.py
```diff
@@ -100,17 +101,32 @@ def __iter__(self):
...
+        if is_dbt:
+            extra_column_values_store = {}
+            extra_columns = self.info_tree.info.tables[0].extra_columns
+            extra_column_diffs = {k:0 for k in extra_columns}
+
         for sign, values in self.result_list:
-            k = values[: len(self.info_tree.info.tables[0].key_columns)]
+            k = values[: len(key_columns)]
+            if is_dbt:
+                extra_column_values = values[len(self.info_tree.info.tables[0].key_columns) :]
```
dlawin, line 118:
```suggestion
                extra_column_values = values[len(key_columns) :]
```
I think this is equivalent

dlawin (approved):
Approving with the performance/readability caveats I've mentioned

data_diff/diff_tables.py
dlawin: If we're using `len(key_columns)` in a couple places we should consider a len_key_columns variable

data_diff/diff_tables.py
leoebfolsom: I agree, it's much more confusing now that the logic isn't duplicated across two methods. I ran it locally (in the context of `data-diff --dbt`) and it seemed pretty quick, so I don't think it's an immediate performance issue.

dlawin, line 104: reminder to run `black -l 120 data_diff/diff_tables.py` if you haven't

data_diff/diff_tables.py
leoebfolsom: @dlawin made this change and created `len_key_columns` 👍

data_diff/diff_tables.py
leoebfolsom: But I definitely want to refactor this to make it the right balance of readable and not-redunant.

PR #385, opened 2023-02-09
Switch database connectors to sqeleton

williebsweet:
Switch the connector docs from data-diff to sqeleton - e.g. `pip install 'data-diff[snowflake]'` to `pip install 'sqeleton[snowflake]'`

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,21 +64,21 @@ pip install data-diff
 
 #### Then, install one or more driver(s) specific to the database(s) you want to connect to.
 
-- `pip install 'data-diff[mysql]'`
+- `pip install 'sqeleton[mysql]'`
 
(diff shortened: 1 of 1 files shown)


erezsh:
I'm not sure this is necessary. The current plan is to support both ways, and the current way is probably less confusing. Also, it allows users to install everything at once, instead of installing data-diff, and then `sqeleton[...]` separately.

williebsweet:
So you can still use `pip install 'data-diff[snowflake]'`?

erezsh:
Yes, should work.

williebsweet:
Confirmed it works. 

However, if you don't have the connector installed the CLI will prompt you to install `'sqeleton[snowflake]'`. Might be good to change that in the future.

Issue #388, opened 2023-02-10
data-diff --dbt --cloud creates accurate, but reversed, diff results


leoebfolsom:
**Describe the bug**
Diff results in the Datafold should be switched--prod should be on the left, dev on the right. Same with the CLI printout.

<img width="946" alt="Screenshot 2023-02-10 at 15 52 31" src="https://user-images.githubusercontent.com/1799931/218223551-0ac052ee-da85-4614-81b9-6548243481d2.png">

<img width="912" alt="Screenshot 2023-02-10 at 15 54 32" src="https://user-images.githubusercontent.com/1799931/218223645-e369f6ec-3aa9-4a3b-a992-05831dd2ef7e.png">

dlawin:
I think this can be resolved easily here: https://github.com/datafold/data-diff/blob/29bfa73105260e680d7769b5dd7bc993c637204a/data_diff/dbt.py#L194

`diff_vars.dev_path` -> table2, `diff_vars.prod_path` -> table1

Assuming that is what determines the order in the UI

dlawin:
Requires an api key to test, so likely a change for Datafold vs. community.

PR #389, opened 2023-02-13
reverse order of dev and prod tables in cli printout and cloud diffs

leoebfolsom:
Addressing https://github.com/datafold/data-diff/issues/388 .

Screenshot 2023-02-13 at 14 21 30

Screenshot 2023-02-13 at 14 22 22

### DIFF
--- a/data_diff/dbt.py
+++ b/data_diff/dbt.py
@@ -65,9 +65,9 @@ def dbt_diff(
         elif is_cloud:
             rich.print(
                 "[red]"
-                + ".".join(diff_vars.dev_path)
-                + " <> "
                 + ".".join(diff_vars.prod_path)
+                + " <> "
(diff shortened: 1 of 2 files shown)


kylemcnair:
Seems like tests need to be updated with the new order too:
https://github.com/datafold/data-diff/blob/master/tests/test_dbt.py#L428
```
        self.assertEqual(request_data_dict["table1"], dev_qualified_list)
        self.assertEqual(request_data_dict["table2"], prod_qualified_list)
```

just the first failure I looked at, didn't go through them all

leoebfolsom:
Thanks @kylemcnair! I think there was only the one failure--I've fixed that (I think) so let's see if CI passes now.

williebsweet (approved):
Approved - pending Dan's review as well

dlawin (approved):

PR #390, opened 2023-02-14
Query optimizer hints support

RoderickJDunn:
Add support for passing `optimizer_hints` to sqeleton through the Python API. 

This depends on sqeleton PR [#19](https://github.com/datafold/sqeleton/pull/19)

Unittests passed after updating the toml file temporarily to point to the above sqeleton branch as follows:

```
[tool.poetry.dependencies]
...
sqeleton = { git = "https://github.com/roderickjdunn/sqeleton.git", rev = "7457caa" }
...
```

### DIFF
--- a/data_diff/__init__.py
+++ b/data_diff/__init__.py
@@ -79,6 +79,8 @@ def diff_tables(
     materialize_all_rows: bool = False,
     # Maximum number of rows to write when materializing, per thread. (joindiff only)
     table_write_limit: int = TABLE_WRITE_LIMIT,
+    # Optimizer hints for Select queries
+    optimizer_hints: Optional[str] = None
 ) -> Iterator:
     """Finds the diff between table1 and table2.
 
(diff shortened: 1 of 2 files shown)


erezsh (commented):
Maybe you should be changing `make_select()` instead?

RoderickJDunn:
You mean so the optimizer hints are used for other SELECT's as well? I added the hint args where I did because the min/max key and checksum SELECT's seem to be by far the slowest queries. didn't think optimizer hints were necessary for `get_values` since its generally quick. Are there other queries I'm missing?

erezsh:
The question is would it do any harm in the rest of the selects? If not, better to have it than not, right?

And this way, if new selects are added, it will be included there too (As long as they use make_select()).

But I don't have a lot of experience with optimizer hints, so maybe my thinking is wrong on this one! Let me know what you think.

RoderickJDunn:
I think it could.. if the hint is a `FULL(table)` hint in Oracle for example, it would force a full-table scan for all select's, even if it's more efficient to use and index-scan on most queries. Maybe the solution is to allow defining separate hints for each query type, but I don't want to complicate the options too much so what do you think?

`query_key_range_hints`: ....
`query_checksum_hints:`: ....

dlawin:
Same thing here 😂 https://github.com/datafold/data-diff/pull/414#issuecomment-1583288204

PR #410, opened 2023-02-24
Unify output formats for `--dbt` running locally

### LINKED ISSUE #409: Make all format options available when using `--dbt`
Currently, the `--dbt` flag uses a different output format than non- `--dbt` executions of `data-diff`.

Ideally, regular `data-diff` and `data-diff --dbt` would have the same output format. Additionally, `--json` and `--stats` should be possible when using `--dbt`.

It's an option for `--dbt` to have completely different formatting and for the `--json` and `--stats` flags to not have any effect. But that alternative makes dolphins cry 🐬 😢 

**Additional context**
I'll submit a PR!

dbeatty10:
resolves #409

## Implementation details

This works by using the same code to render the diff output. It does this for two major cases:
1. When `--dbt` is not used
1. When `--dbt` (and `not is_cloud`)

The one case that still gets it's own output is `--dbt` (and `is_cloud`).

I manually tested the first two cases, but I did not test the 3rd case.

### Default
```shell
$ data-diff --dbt
```

```
Found 1 successful model runs from the last dbt command.
- 1, black
+ 1, orange
```

image

### JSON
```shell
$ data-diff --dbt --json 
```

```shell
Found 1 successful model runs from the last dbt command.
["-", ["1", "black"]]
["+", ["1", "orange"]]
```

image

### Stats
```shell
$ data-diff --dbt --stats
```

```shell
Found 1 successful model runs from the last dbt command.
2 rows in table A
2 rows in table B
0 rows exclusive to table A (not present in B)
0 rows exclusive to table B (not present in A)
1 rows updated
1 rows unchanged
50.00% difference score

Extra-Info:
  diff_counts = {'id_a': 0, 'color_a': 0}
  exclusive_count = 0
  table1_count = 2
  table2_count = 2
  validated_unique_keys = [['id'], ['id']]
```

image

### DIFF
--- a/data_diff/__main__.py
+++ b/data_diff/__main__.py
@@ -259,11 +259,13 @@ def main(conf, run, **kw):
 
     try:
         if kw["dbt"]:
-            dbt_diff(
+            diff = dbt_diff(
                 profiles_dir_override=kw["dbt_profiles_dir"],
                 project_dir_override=kw["dbt_project_dir"],
                 is_cloud=kw["cloud"],
             )
+            render_diff(diff, kw["limit"], kw["stats"], kw["json_output"])
+
         else:
             return _data_diff(**kw)
     except Exception as e:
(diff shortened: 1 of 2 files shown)


dlawin:
Hey @dbeatty10, apologies that we let this PR sit so long. I do think it makes sense to have output formats available across all execution methods, at the same time there has been a considerable amount of branch drift since this change was opened, so I'm thinking we should close this PR.

What do you think?

Issue #401, opened 2023-02-19
Unexpected behavior using min_key/max_key


RoderickJDunn:
**Describe the bug**
When I pass in the `min_key` and/or `max_key` parameters, I would expect data-diff to skip the `query_key_range` step since it already knows the min/max keys. 

For example, if I want to restrict the account ID range by passing in `min_key=000000000001` and `max_key=R90909RDD099`, the tool still executes `query_key_range` with a query like this:
```
SELECT cast(min("ACCT_ID") as varchar(1024)), cast(max("ACCT_ID") as varchar(1024)) 
FROM "accounts" 
WHERE ("ACCT_ID" >= '000000000001') AND ("ACCT_ID" < 'R90909RDD099')
```

IMO this query should've been skipped since the user already told data-diff the min/max values for ACCT_ID.

This query can be slow for large tables, so it would be great to bypass it if the min/max keys are already known.

**Describe the environment**
Ubuntu
data-diff version: [608312d944ada6ba07c810f25a4bd638933c8fc1](https://github.com/datafold/data-diff/commit/608312d944ada6ba07c810f25a4bd638933c8fc1)

erezsh:
I agree. We should only query for these values if they are set to None.

PR #414, opened 2023-02-24
Key Range Query adjustments

RoderickJDunn:
**Changes**
- If `min_key` + `max_key` are both passed in as parameters, the `query_key_range` query is bypassed
- If only one of `min_key` or `max_key` are passed in, `query_key_range` is called but the min or max is overridden with the user-provided value.

fixes [#401](https://github.com/datafold/data-diff/issues/401)

### DIFF
--- a/data_diff/diff_tables.py
+++ b/data_diff/diff_tables.py
@@ -60,9 +60,9 @@ class ThreadBase:
             for future in as_completed(futures):
                 yield future.result()
 
-    def _threaded_call_as_completed(self, func, iterable, *args, **kwargs):
+    def _threaded_call_as_completed(self, func, iterable):
         "Calls a method for each object in iterable. Returned in order of completion."
-        return self._thread_as_completed(methodcaller(func, *args, **kwargs), iterable)
+        return self._thread_as_completed(methodcaller(func), iterable)
 
     @contextmanager
     def _run_in_background(self, *funcs):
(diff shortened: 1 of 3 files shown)


data_diff/diff_tables.py
```diff
@@ -290,8 +290,21 @@ def _bisect_and_diff_tables(self, table1, table2, info_tree):
         if key_type.python_type is not key_type2.python_type:
             raise TypeError(f"Incompatible key types: {key_type} and {key_type2}")
 
-        # Query min/max values
-        key_ranges = self._threaded_call_as_completed("query_key_range", [table1, table2])
+        if all([table1.min_key, table1.max_key, table2.min_key, table2.max_key]):
```
erezsh, line 293: You should test for `is None`, because keys can be 0.

erezsh (changes requested):
This is too complicated.

I don't think you need to change anything except for `_bisect_and_diff_tables()`. And possibly change `query_key_range()` signature to `query_key_range(self, get_min=True, get_max=True)`. (though always querying both min and max, and discarding the one you don't need, is simpler, and won't affect performance in a meaningful way)

data_diff/diff_tables.py
RoderickJDunn: Ah true, will fix that

RoderickJDunn:
Ok I'll simplify things :)

data_diff/diff_tables.py
```diff
@@ -271,6 +271,14 @@ def _diff_segments(
     ):
         ...
 
+    def _resolve_key_range(self, key_range_res, usr_key_range):
```
RoderickJDunn, line 274: extracted to a function since this logic is needed twice.

dlawin:
Hey @RoderickJDunn, apologies that this PR has been sitting so long. Is this a change you're still interested in making?

dlawin:
Sounds good @RoderickJDunn, let us know!

PR #415, opened 2023-02-25
Use the same logic as dbt-core for the path for the project and profiles

### LINKED ISSUE #416: Use the same logic as dbt-core or the paths for the project and profiles
If I have a dbt `profiles.yml` in my current working directory, I'd like to be able to do this:

```shell
data-diff --dbt
```

This is the way to do it currently:
```shell
data-diff --dbt --dbt-profiles-dir .
```

**Additional context**
I'll open a PR for this.

dbeatty10:
resolves #416

### Overview

The changes in this PR allow using the same logic as dbt. The precedence for the location of `profiles.yml` changed starting dbt-core>=1.4.0.

This PR is designed to determine the version of dbt indicated within the manifest and use the applicable logic.

#### Behavior for dbt >= 1.3

The parent directory for profiles.yml is determined using the following precedence:

1. `--profiles-dir` option (`--dbt-profiles-dir` in `data-diff`)
1. `DBT_PROFILES_DIR` environment variable
1. current working directory **← NEW**
1. `~/.dbt/` directory

#### Behavior prior to 1.3:

The parent directory for profiles.yml is determined using the following precedence:

1. `--profiles-dir` option (`--dbt-profiles-dir` in `data-diff`)
1. `DBT_PROFILES_DIR` environment variable
1. `~/.dbt/` directory

Both are documented [here](https://docs.getdbt.com/docs/get-started/connection-profiles#advanced-customizing-a-profile-directory). The version needs to be changed at the top of the screen to switch between versions.

image

### Implementation details

[`Path`](https://docs.python.org/3/library/pathlib.html) is used liberally.

### Source of the logic

The relevant logic is located here in `dbt-core`:
- https://github.com/dbt-labs/dbt-core/blob/c952d44ec5c2506995fbad75320acbae49125d3d/core/dbt/cli/resolvers.py#L6-L13


erezsh (approved):

tests/test_dbt.py
```diff
@@ -54,7 +54,7 @@ def test_get_datadiff_variables_empty(self):
     def test_get_models(self, mock_manifest_parser, mock_run_parser, mock_open):
         expected_value = "expected_value"
         mock_self = Mock()
-        mock_self.project_dir = ""
+        mock_self.project_dir = Path()
```
dlawin, line 57: I think we're missing an import

```
NameError: name 'Path' is not defined
```

dbeatty10: Thanks @dlawin -- just pushed a commit that adds the import.

dbeatty10:
I'm getting this error, so I'll just need to look at the way the test is expressed and update the test accordingly:
```
Expected: open('target/run_results.json')
Actual: open(PosixPath('target/run_results.json'))
```

Alternatively, I could cast all the `Path`s to `str`s and the tests should pass. But it seems like using a `Path` in `open` should be acceptable.

dlawin:
> 



> I'm getting this error, so I'll just need to look at the way the test is expressed and update the test accordingly:
> 
> ```
> Expected: open('target/run_results.json')
> Actual: open(PosixPath('target/run_results.json'))
> ```
> 
> Alternatively, I could cast all the `Path`s to `str`s and the tests should pass. But it seems like using a `Path` in `open` should be acceptable.

yeah I should have used `Path` originally

dlawin:
This is dbt's behavior only as of >=1.4 correct?

Just we're we cause a disconnect for dbt < 1.4

data_diff/dbt.py
```diff
@@ -31,14 +32,26 @@ def import_dbt():
...
+RUN_RESULTS_PATH = "target/run_results.json"
+MANIFEST_PATH = "target/manifest.json"
+PROJECT_FILE = "dbt_project.yml"
+PROFILES_FILE = "profiles.yml"
 LOWER_DBT_V = "1.0.0"
 UPPER_DBT_V = "1.5.0"
 
 
+# https://github.com/dbt-labs/dbt-core/blob/c952d44ec5c2506995fbad75320acbae49125d3d/core/dbt/cli/resolvers.py#L6
+def default_project_dir() -> Path:
```
dlawin, line 44: we should update the help text here https://github.com/datafold/data-diff/blob/d2d78495e650fd336c762f164315cb7e1b706074/data_diff/__main__.py#L208

```
      --dbt-profiles-dir PATH         Override the default dbt profile location
                                      (~/.dbt).
      --dbt-project-dir PATH          Override the dbt project directory.
                                      Otherwise assumed to be the current
                                      directory.
```

dbeatty10: Great catch @dlawin !

Added updates to the help text in commit https://github.com/datafold/data-diff/pull/415/commits/098565af43e7994b900e31e5e1ea594b78644866.

Copied verbatim from [here](https://github.com/dbt-labs/dbt-core/blob/0ef9931d197c996ccd4e3caf641729176eceda99/core/dbt/cli/params.py#L243-L257). Can adjust the language if you want.

dlawin:
> This is dbt's behavior only as of >=1.4 correct?
> 
> Just worried we cause a disconnect for dbt < 1.4

I think we should make this change and align with dbt, but I do think we should also consider how to handle <1.4

The shift of this default location was a little jarring for dbt users, and anecdotally I've seen many work around the change via the environment variable `export DBT_PROFILES_DIR=~/.dbt/`

If we had a `data-diff --dbt-debug` feature analogous to `dbt debug`, or were a little more verbose when starting a `data-diff --dbt`, I'd feel a little more comfortable about it.

The location of this file is a little chaotic at the moment, and the <dbt_project>/profiles.yml file will often have `{{env_var()}}`s that are not set on local dev environments. With `--dbt`'s current state, there could be some very obscure errors. Ironically, I think those on 1.4+ are more aware of this location being a potential issue than those who aren't

dbeatty10:
Agreed that it makes sense to align with the behavior based on the dbt version.

I just pushed some changes to try handling both the `DBT_PROFILES_DIR` environment variable as well as the change in behavior between dbt 1.2 and dbt 1.3.

⚠️ I did **not** test them locally yet though.

#### Behavior for dbt >= 1.3

The parent directory for profiles.yml is determined using the following precedence:

1. `--profiles-dir` option
1. `DBT_PROFILES_DIR` environment variable
1. current working directory
1. `~/.dbt/` directory

#### Behavior prior to 1.3:

The parent directory for profiles.yml is determined using the following precedence:

1. `--profiles-dir` option
1. `DBT_PROFILES_DIR` environment variable
1. `~/.dbt/` directory

Both are documented [here](https://docs.getdbt.com/docs/get-started/connection-profiles#advanced-customizing-a-profile-directory). The version needs to be changed at the top of the screen to switch between versions.

image

dlawin (approved):
Works great, really appreciate the contributions Doug!

dbeatty10:
Like the wording update @dlawin -- thanks for putting that in there 🏅

williebsweet:
Thanks, @dbeatty10!

Issue #412, opened 2023-02-24
Make the DBT dependency optional


erezsh:
Alternative title: Sudden dependency bloat in data-diff 4.0 and up

# Overview of the problem

Version 0.3.2 of data-diff required 11 dependencies total, most of which are already very popular libraries, like "rich", "click", and "dsnparse".

Version 0.4.0 of data-diff introduced 66 new dependencies (!).

This is due to the introduction of the mandatory "dbt" dependency, even for users who don't plan to use the `--dbt` switch.

# Implications of not fixing this

1. Minimum load time for the tool increased from `0.165` seconds to `2.410` seconds (!!)
  
2. Extra requirements might collide with the requirements of other Python libraries, and make it harder for our non-dbt users to have data-diff installed alongside them.

3. Users who would consider using data-diff as a lightweight tool might be put out by the large number of dependencies.

4. That is also true in regards to being included by default in package managers. For example, in Ubuntu you can `apt install python3-lark` to install Lark. It would have been much harder to include it if it had many dependencies.

# Implications of fixing this

Users that don't already have dbt installed, and run `--dbt`, will see a message telling them to use `pip install data-diff[dbt]` which will install dbt.

However, there is absolutely no point in using the `--dbt` switch if you don't already have dbt installed and configured.

# Conclusion

I think there is no reason to keep dbt as a dependency, but very good reasons to make it optional.

dlawin:
it may be possible to remove the dependency, the dbt package is only used for one method here:
https://github.com/datafold/data-diff/blob/584206b407320a36fc95a204b50e09f8d20239a8/data_diff/dbt.py#L311

to render dbt-jinja {{ env_vars('')}}

williebsweet:
@erezsh Thoughts on making this part of a guided setup?

In this scenario, someone installs data-diff, then is asked a series of questions that will install relevant dependencies for them, e.g., which databases do they use, do they use dbt, if dbt - what is their prod db and schema, etc.

erezsh:
@williebsweet Generally, it's a good idea, but it belongs in a separate issue.

In this particular case, I'll reiterate my opinion -

> there is absolutely no point in using the --dbt switch if you don't already have dbt installed and configured.

dlawin:
I noticed numpy was still in the poetry.lock file when incrementing the duckdb version here https://github.com/datafold/data-diff/pull/408/commits/b939b9a0ab991c8a452be3ca4c73ab2958178e6a

It's removed now, so that should help with this effort

PR #421, opened 2023-02-27
Make DBT dependency optional

erezsh:
Fixes Issue #412

### DIFF
--- a/data_diff/dbt.py
+++ b/data_diff/dbt.py
@@ -3,7 +3,6 @@ import logging
 import os
 import time
 import rich
-import yaml
 from dataclasses import dataclass
 from packaging.version import parse as parse_version
 from typing import List, Optional, Dict
@@ -14,10 +13,11 @@ def import_dbt():
     try:
         from dbt_artifacts_parser.parser import parse_run_results, parse_manifest
         from dbt.config.renderer import ProfileRenderer
+        import yaml
     except ImportError:
         raise RuntimeError("Could not import 'dbt' package. You can install it using 'pip install data-diff[dbt]'.")
 
-    return parse_run_results, parse_manifest, ProfileRenderer
+    return parse_run_results, parse_manifest, ProfileRenderer, yaml
 
 from .tracking import (
     set_entrypoint_name,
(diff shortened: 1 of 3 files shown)


erezsh:
@dlawin Any idea why these tests are failing?

dlawin:
> @dlawin Any idea why these tests are failing?

Hopefully it's clear from this: [21804e5](https://github.com/datafold/data-diff/pull/421/commits/21804e5d9553e2d1af32e28ba71b3903102b6411)

Need to use the mock_self as the mock, since that's how the methods are being called now. No need to mock the package

dlawin:
> > @dlawin Any idea why these tests are failing?
> 
> Hopefully it's clear from this: [21804e5](https://github.com/datafold/data-diff/pull/421/commits/21804e5d9553e2d1af32e28ba71b3903102b6411)
> 
> Need to use the mock_self as the mock, since that's how the methods are being called now. No need to mock the package (and a package level mock won't override the self.blah methods now)

I can fix the remaining ones later today

data_diff/dbt.py
```diff
@@ -3,14 +3,21 @@
...
-from dbt_artifacts_parser.parser import parse_run_results, parse_manifest
-from dbt.config.renderer import ProfileRenderer
+
+def import_dbt():
+    try:
+        from dbt_artifacts_parser.parser import parse_run_results, parse_manifest
+        from dbt.config.renderer import ProfileRenderer
+        import yaml
+    except ImportError:
+        raise RuntimeError("Could not import 'dbt' package. You can install it using 'pip install data-diff[dbt]'.")
```
williebsweet, line 18: Does this need to be `pip install 'data-diff[dbt]'`?

erezsh:
@dlawin Nice, thank you!

data_diff/dbt.py
williebsweet: I disagree - we should address this particular case in this particular issue. No reason to create an error that is not useful.

We can address the other instances in follow up issues but we should address this one in this issue.

erezsh: So what do you suggest? There is no single form of writing it that will work for everyone.

williebsweet: Does `pip install 'data-diff[dbt]'` work for some people and `pip install 'data-diff[dbt]'` work for others?

williebsweet: I don't understand the pushback. I know we have an issue though. If I run `pip install data-diff[dbt]` - I get the following "no matches found: data-diff[dbt]". Which may leave users confused and without a next step.

erezsh: Done.

williebsweet (approved):

PR #446, opened 2023-03-17
WIP Add support for inferring PKs via tests

### LINKED ISSUE #439: Infer PKs via uniqueness tests
Having to tag every PK column with "primary-key" is an annoying amount of setup, especially when it's easy to infer the model's PKs from dbt tests without additional configuration for `data-diff`

Infer the PK column(s) from uniqueness tests when no "primary-key" tag is present

We should consider including the primary key(s) in the output so that an incorrect assumption is more obvious

dlawin:
Resolves #439


data_diff/dbt.py
```diff
@@ -105,7 +109,7 @@ def dbt_diff(
                 + " <> "
                 + ".".join(diff_vars.dev_path)
                 + "[/] \n"
-                + "Skipped due to missing primary-key tag(s).\n"
+                + "Skipped due to unknown primary key. Add uniqueness tests, meta, or tags.\n"
```
dlawin, line 112: could probably use some wordsmithing

data_diff/dbt.py
```diff
@@ -315,24 +320,24 @@ def get_models(self):
...
-            manifest_obj = self.parse_manifest(manifest=manifest_dict)
-
         success_models = [x.unique_id for x in run_results_obj.results if x.status.name == "success"]
-        models = [manifest_obj.nodes.get(x) for x in success_models]
+        models = [self.manifest_obj.nodes.get(x) for x in success_models]
         if not models:
             raise ValueError("Expected > 0 successful models runs from the last dbt command.")
 
-        rich.print(f"Found {str(len(models))} successful model runs from the last dbt command.")
+        rich.print(f"Found {str(len(models))} successful model runs from the last dbt command. Starting diffs.\n")
```
dlawin, line 328: words?

kylemcnair: I agree, it feels awkward as a user to be left hanging when diffs aren't instant.

Looking for additional options, dbt does "Running with dbt=1.4.0" after a run command, maybe "Running with data-diff={version}"

williebsweet: That's a good idea @kylemcnair.

It'll also reinforce what version they are on so when we upgrade and talk about it, they'll have a better idea of whether they can access the newest features.

data_diff/dbt.py
kylemcnair: I think you're explaining the error and giving actionable feedback, so the message seems ok to me.

Couple ideas if you're looking for options or ideas

> Skipped due to unknown primary key. Model primary keys must have a uniqueness test, or  have a "primary-key" value in the meta or tag configuration.

> Skipped due to unknown primary key. Help data-diff infer the primary key by adding uniqueness tests, meta, or tags.

dlawin: I think I will leave it. Always a trade off between brevity/info

data_diff/dbt.py
dlawin: I like it

dlawin: updated

PR #458, opened 2023-03-25
Various improvements and features

RoderickJDunn:
### Summary
- Support overriding column types
- Added Grouping Hash algorithm to support single-pass hashing on large Oracle tables. 
- Disabled tracking
- Support sorting of results when single column of composite PK is used to find hashing ranges
- Support early abort (allows us to exit if timeout or result limit is reached)
- **Support custom column conversions**

## Supporting custom column conversions
Today data-diff always preprocesses columns in the same way for a given data type. This allows for normalization of columns between databases, so that hash comparisons are valid. 

However, there are some cases where we need to specify custom processing of a column in order to create a hash that can be compared to that of the other database. For example the `xxbrk_daily_price` Oracle table contains several columns of type `number` with no precision or scale specified. Fivetran may choose to replicate such columns to Redshift as strings. Today, for this case data-diff will fail right away due to a column type mismatch (it can't compare numerics with strings).

### Why simple casting doesn't work
- I tried to overcome the above issue by supporting column type overrides, but this feature is inadequate in this case. 
- The Oracle columns contain values with very wide range of scales - eg. values of both `1` and `99.01234567890123456789012345678901234567` are possible (that's a scale of 38).
- Those values also exist in Redshift, but they are strings
- If I treat both columns as numeric(38,10), most comparisons pass, but many fail due to minute differences caused by rounding. In Oracle, casting to number(38,10) will yield a rounded number, but in Redshift casting the string to numeric(38,10) will truncate the number.
- If I treat both columns as strings, this similarly fails because Oracle implicitly rounds values before casting to varchar.  

### Solution
- Custom processing of columns allows us to TRUNCATE the oracle values, and thus prevent the implicit rounding from occurring. In summary:
  - Oracle columns are TRUNCATED to a scale of 10, then converted to a string with a normalized format
  - Redshift columns are cast to numeric(38,10), then converted to a string with the same normalized format

### DIFF
--- a/data_diff/__init__.py
+++ b/data_diff/__init__.py
@@ -81,7 +81,9 @@ def diff_tables(
     # Maximum number of rows to write when materializing, per thread. (joindiff only)
     table_write_limit: int = TABLE_WRITE_LIMIT,
 
-    hash_query_type: str = None
+    hash_query_type: str = None,
+    # Optimizer hints for Select queries
+    optimizer_hints: Optional[str] = None
 ) -> Iterator:
     """Finds the diff between table1 and table2.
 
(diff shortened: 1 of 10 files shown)


RoderickJDunn:
Apologies this was not meant to be opened here yet. Feel free to delete

erezsh:
You can force-push and re-open when it's ready.

PR #462, opened 2023-03-28
cloud api token flow

pik94:
Add an option of generating of a Datafold API key to interact with a datafold cloud.

If the key is not provided, it is suggested to a user to generate it redirecting to a corresponding datafold cloud page.
Also add an opportunity to set a datafold host (by default it is our cloud url https://app.datafold.com)

### DIFF
--- a/data_diff/dbt.py
+++ b/data_diff/dbt.py
@@ -232,7 +232,7 @@ def _cloud_diff(diff_vars: DiffVars) -> None:
             webbrowser.open(f"{datafold_host}/login?next={datafold_host}/users/me")
             return
         else:
-            raise ValueError('Cannot diff because the API key is not provided')
+            raise ValueError("Cannot diff because the API key is not provided")
 
     if diff_vars.datasource_id is None:
         raise ValueError(
--- a/data_diff/dbt.py
+++ b/data_diff/dbt.py
@@ -269,7 +269,6 @@ def _cloud_diff(diff_vars: DiffVars) -> None:
         response.raise_for_status()
         data = response.json()
         diff_id = data["id"]
-        # TODO in future we should support self hosted datafold
         diff_url = f"{datafold_host}/datadiffs/{diff_id}/overview"
         rich.print(
             "[red]"
--- a/data_diff/dbt.py
+++ b/data_diff/dbt.py
@@ -1,7 +1,9 @@
 import json
 import os
 import time
+import webbrowser
 import rich
+from rich.prompt import Confirm
 
 from collections import defaultdict
 from dataclasses import dataclass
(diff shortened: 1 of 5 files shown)


data_diff/dbt.py
```diff
@@ -256,7 +270,7 @@ def _cloud_diff(diff_vars: DiffVars) -> None:
         data = response.json()
         diff_id = data["id"]
         # TODO in future we should support self hosted datafold
```
dlawin, line 272: this TODO can be removed

data_diff/dbt.py
```diff
@@ -220,16 +222,28 @@ def _local_diff(diff_vars: DiffVars) -> None:
...
 def _cloud_diff(diff_vars: DiffVars) -> None:
+    datafold_host = os.environ.get("DATAFOLD_HOST")
+    if datafold_host is None:
+        datafold_host = "https://app.datafold.com"
+    datafold_host = datafold_host.rstrip("/")
+    rich.print(f"Cloud datafold host: {datafold_host}")
+
     api_key = os.environ.get("DATAFOLD_API_KEY")
+    if not api_key:
+        rich.print("[red]API key not found, add it as an environment variable called DATAFOLD_API_KEY.")
```
nolar, line 233: This might be inconvenient to set the env var in every terminal tab. Putting the token in the standard shell setup scripts might (and will) compromise the security.

It is better to switch to using [`keyring`](https://pypi.org/project/keyring/) sooner (maybe not in this PR, but fast enough).

nolar (approved):

data_diff/dbt.py
```diff
@@ -220,16 +222,28 @@ def _local_diff(diff_vars: DiffVars) -> None:
 
 
 def _cloud_diff(diff_vars: DiffVars) -> None:
+    datafold_host = os.environ.get("DATAFOLD_HOST")
+    if datafold_host is None:
+        datafold_host = "https://app.datafold.com"
+    datafold_host = datafold_host.rstrip("/")
+    rich.print(f"Cloud datafold host: {datafold_host}")
```
dlawin, line 229: Not sure about printing this every time. Might make more sense as a debug log

data_diff/dbt.py
pik94: I think we should support env vars too by default because it is a common wide-spreaded way.
I am planning to add [keyring](https://pypi.org/project/keyring/) in the next iteration although i have read that it might be some problem to install in on Linux. Anyway, i will try.

data_diff/dbt.py
pik94: I add it just in case in order to a user can see that the correct host is used.

PR #465, opened 2023-03-28
override --cloud host name

### LINKED ISSUE #461: Support a custom endpoint for --cloud
The endpoint for `--cloud` diffs is hardcoded to app.datafold.com, add a way to specify a custom endpoint and support on-prem datafold customers

A flag could work, but an environment variable or maybe a variable in the dbt_project would simplify the cli command

dlawin:
Resolves #461 

Adds an additional flag `--cloud-host-name`. This flag or the env variable `DATAFOLD_API_HOST_NAME` can be used to override the url for `--cloud`

Example: 
`--cloud-host-name blah.blah.com` -> "blah.blah.com/api/v1/datadiffs" being used. I don't think there's a reason to override the full endpoints

### DIFF
--- a/data_diff/dbt.py
+++ b/data_diff/dbt.py
@@ -73,7 +73,10 @@ class DiffVars:
 
 
 def dbt_diff(
-    profiles_dir_override: Optional[str] = None, project_dir_override: Optional[str] = None, is_cloud: bool = False
+    profiles_dir_override: Optional[str] = None,
+    project_dir_override: Optional[str] = None,
+    is_cloud: bool = False,
+    cloud_host_name: Optional[str] = None,
 ) -> None:
     set_entrypoint_name("CLI-dbt")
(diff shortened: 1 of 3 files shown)


data_diff/dbt.py
```diff
@@ -229,7 +233,7 @@ def _cloud_diff(diff_vars: DiffVars) -> None:
     if api_key is None:
         raise ValueError("API key not found, add it as an environment variable called DATAFOLD_API_KEY.")
 
-    url = "https://app.datafold.com/api/v1/datadiffs"
+    url = f"https://{host_name}/api/v1/datadiffs"
```
dlawin, line 236: @nolar do you see any reason to override the full endpoint urls? I made an assumption that the host_name is enough

dlawin:
Closing as this is addressed in #462

Issue #442, opened 2023-03-15
--dbt add snowflake key pair authentication support


mimoyer21:
**Is your feature request related to a problem? Please describe.**
I am not able to use the --dbt functionality with snowflake key pair authentication
![image](https://user-images.githubusercontent.com/63069704/225440451-8380e319-93fc-45e0-9fba-c947f2e637d5.png)

**Describe the solution you'd like**
Support the connection method described here: https://docs.getdbt.com/reference/warehouse-setups/snowflake-setup#key-pair-authentication

**Describe alternatives you've considered**
The only current option is to use password authentication

**Additional context**
@dlawin asked me to create this issue in [this Slack thread](https://getdbt.slack.com/archives/C03D25A92UU/p1678912218963129?thread_ts=1677252291.169159&cid=C03D25A92UU)

smitsrr:
Looking also for SSH key authentication method for snowflake.

leoebfolsom:
A [similar question has been asked](https://getdbt.slack.com/archives/C03D25A92UU/p1679992566566249) about Redshift. Should we aggregate all requests for other authentication types in this issue, or do we need separate issues for each warehouse? @dlawin @williebsweet  🙏

williebsweet:
> A [similar question has been asked](https://getdbt.slack.com/archives/C03D25A92UU/p1679992566566249) about Redshift. Should we aggregate all requests for other authentication types in this issue, or do we need separate issues for each warehouse? @dlawin @williebsweet 🙏

Open a new one for each db.

PR #468, opened 2023-03-31
add basic snowflake key/pair support

dlawin:
Partial resolution for #442 

Supports [this ](https://docs.getdbt.com/reference/warehouse-setups/snowflake-setup#key-pair-authentication)auth method except when `private_key_passphrase` is provided


kylemcnair (approved):
makes sense to me

Issue #447, opened 2023-03-17
Advanced custom schemas (dynamic schema behavior based on target)


dlawin:
**Is your feature request related to a problem? Please describe.**
It's fairly common to override the `generate_schema_name` macro to control schema names per environment.

For example, you may override it so that you use a single schema for local development, but models write to multiple schema when env == 'prod':

Single local dev schema (<target_schema>)
- dan_dev

Prod schemas (<custom_schema>)
- marketing
- sales
- etc

This differs from the default behavior of dbt, where schema names are constructed as <target_schema>_<custom_schema>

**Describe the solution you'd like**
Ideally this is configurable

Maybe something like:
datadiff: 
  prod_schema: <custom_schema> <---- makes it dynamic. <custom_schema> if exists, otherwise <prod_schema>

Or similarly easy way to configure

https://docs.getdbt.com/docs/build/custom-schemas#changing-the-way-dbt-generates-a-schema-name

Alternatives considered:
Outside of configuration values, we would need a copy of the production manifest.json

brown5628:
+1, supporting the [alternative pattern for generating schema names from the dbt docs](https://docs.getdbt.com/docs/build/custom-schemas#an-alternative-pattern-for-generating-schema-names) would cover the most common non-default way I've seen this done. 

This is currently an adoption blocker for me, and even the simple dynamic example you sketched or the ability to only use the `custom_schema` would give me what I need to move forward!

dlawin:
@brown5628 could you review the description on https://github.com/datafold/data-diff/pull/498 ? Want to verify that will work for you

brown5628:
@dlawin Yes, this would work perfectly. Thank you!

dlawin:
> @dlawin Yes, this would work perfectly. Thank you!

Awesome, let me know if you encounter any issues or if the doc is unclear [here](https://docs.datafold.com/development_testing/open_source/)

This change is available as of data-diff 0.7.0

PR #473, opened 2023-04-03
Refactor: Add Function Use_Prod_Schema

CoreyGaunt:
The `use_prod_schema` is a function that is set to true as a default, but enables users to optionally prune the target schema from schema creation in their prod environments.


dlawin:
This would cover the most common scenario in #447 

Prod schema:
`<custom_schema_name>`

Dev schema:
`<dev_target_schema>_<custom_schema_name>`

Thanks @CoreyGaunt

dlawin:
@CoreyGaunt I'm going to close this for now as I think we can cover every scenario in https://github.com/datafold/data-diff/pull/498

Could you take a look at that?

PR #480, opened 2023-04-05
sqeleton vendoring PoC

vvkh:
Basically
```
mkdir vendor
cd vendor
git clone git@github.com:datafold/sqeleton.git
rm -rf sqeletion/.git
rm -rf sqeleton/.github
```
\+ updates in the pyproject.toml file


erezsh:
What is the point of this change?

vvkh:
@erezsh we briefly discussed the idea with @wvanbergen with the primary goal of improving time to market and simplifying the release process. For now, it's a PoC to check if such setup works without heavy refactoring, so you can ignore the PR.

vvkh:
- [ ] no merge

wvanbergen:
Because we shelved our plans for the other tool that would use the sqeleton library, it doesn't really make much sense to have it in a separate repository anymore. It makes contributing to datadiff harder than it has to be, and makes releases a more complicated process. So we decided to merge these repositories back into a single codebase.

nolar (changes requested):
Thank you for taking care of it. Luckily, we have no rush on this merging, so I suggest that we take a different route — the one that preserves the git history of changes and embeds sqeleton as one extra top-level package into the library — instead of bringing it as a vendored subdirectory which requires import-path injections.

Essentially, that is `git remote add sqeleton …` + `git merge sqeleton/master  --allow-unrelated-histories`, and then resolve the conflicts on the same-named files (mostly CI & metadata). I am already working on it.

pyproject.toml
```diff
@@ -29,7 +29,7 @@ dsnparse = "*"
 click = "^8.1"
 rich = "*"
 toml = "^0.10.2"
-sqeleton = "0.0.8"
+sqeleton = { path = "./vendor/sqeleton", develop = true }
```
nolar, line 32: On a side note: I wonder, will this approach also package the files properly into tarballs & wheels and install it in a clean environment that is not git-based (i.e. `pip install data-diff`)?

nolar:
Here is an alternative proposal with the git history preserved: https://github.com/datafold/data-diff/pull/485

williebsweet:
@nolar @wvanbergen  Should we close this?

PR #482, opened 2023-04-05
allow pass: key for redshift

### LINKED ISSUE #481: pass: as an alternate key for redshift password
Password seems that it can also be specified as `pass:` for redshift

See https://github.com/datafold/data-diff/pull/382#issuecomment-1497672929

dlawin:
Resolves #481

### DIFF
--- a/data_diff/dbt.py
+++ b/data_diff/dbt.py
@@ -324,12 +324,7 @@ def _cloud_diff(diff_vars: DiffVars, datasource_id: int, datafold_host: str, url
             send_event_json(event_json)
 
         if error:
-            rich.print(
-                "[red]"
-                + ".".join(diff_vars.prod_path)
-                + " <> "
-                + ".".join(diff_vars.dev_path) + "[/]\n"
-            )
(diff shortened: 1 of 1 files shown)


data_diff/dbt.py
```diff
@@ -324,12 +324,7 @@ def _cloud_diff(diff_vars: DiffVars, datasource_id: int, datafold_host: str, url
             send_event_json(event_json)
 
         if error:
-            rich.print(
-                "[red]"
-                + ".".join(diff_vars.prod_path)
-                + " <> "
-                + ".".join(diff_vars.dev_path) + "[/]\n"
-            )
+            rich.print("[red]" + ".".join(diff_vars.prod_path) + " <> " + ".".join(diff_vars.dev_path) + "[/]\n")
```
williebsweet, line 327: Should we use an f-string to keep it consistent with the print statements below?

williebsweet (approved):

data_diff/dbt.py
dlawin: yes I'll fix that in this PR https://github.com/datafold/data-diff/pull/464/files

dlawin: done

PR #485, opened 2023-04-06
Embed sqeleton with git history preserved

nolar:
We modify & release `sqeleton` & `data-diff` in sync quite often. The latter strictly depends on newer versions of the former. However, the former (sqeleton) is not widely adapted as a standalone library, so this process is rather a burden and does not bring benefits. Also see https://github.com/datafold/data-diff/pull/480#issuecomment-1497997704

It would be easier to modify the source code & release the data-diff library in one step. So, we merged data-diff & sqeleton for now. But we keep the possibility of their splitting indefinitely later. 

`sqeleton` remains a top-level package for backward compatibility with the existing code (e.g. `import sqeleton`).

The git history & code authorship are preserved. Sqeleton's git tags are lost — not to cause conflict with data-diff's tags. However, the historic version numbers are available in the code. Sqeleton's metadata is anyway removed after the merge.

**👀Reviewers:**  There are too many commits. Only look at the latest one where the merge happens.

**TODOs:**

- [ ] ❓ How do we uninstall `sqeleton` from the existing venvs and pinned requirements — to avoid the package name conflict?
  - _(The fallback plan: rename it or move it to `data_diff.sqeleton`. The only drawback is a lengthy package name.)_

---

Essentially, this was done:

```
git remote add sqeleton git@github.com:datafold/sqeleton.git
git fetch sqeleton

git checkout -b embed-sqeleton
git merge --allow-unrelated-histories sqeleton/master 
```

Then resolve all the conflicts, shuffle the files in their appropriate places (CI, docs, package metadata), cross-link then as needed, commit and push.


wvanbergen:
Needs some conflict resolution, but let's get this done ASAP.

nolar:
This force-push is a replay of the same changes on top of the newer main branches in both repos.

Also, sqeleton docs are removed. The TUI for "connection editor" is also removed to resolve the package conflict for Textual (see the PR description).

vvkh (approved):

wvanbergen:
Thanks @vvkh and @nolar for making this happen.

PR #491, opened 2023-04-11
Grab credentials from profiles.yml

pik94:
Provide an opportunity to users to extract database credentials from dbt `profiles.yml` when creating a new data source in the Datafold cloud.

When the user reaches the step of entering database credentials when creating a new data source in the Datafold cloud, he is prompted to use the credentials from `profiles.yml`. If he gives his consent to this, we collect the necessary data from the file. 

If for some reason we were unable to extract this data from the file, a corresponding message is displayed in the terminal and it is suggested to enter this data manually.


data_diff/dbt_parser.py
nolar, line 1: Can you please highlight which code is new here? The side-by-side comparison shows no noticeable differences. Or is it just an extraction into a separate file with no actual changes?

nolar (approved):

data_diff/dbt_parser.py
pik94: Actually, there are no new code there. I just moved `DbtParser` to a separate module to avoid cyclic imports.

data_diff/cloud/data_source.py
```diff
@@ -64,7 +68,25 @@ def create_ds_config(ds_config: TCloudApiDataSourceConfigSchema, data_source_nam
...
-def _parse_ds_credentials(ds_config: TCloudApiDataSourceConfigSchema, only_basic_settings: bool = True):
+def _parse_ds_credentials(
+    ds_config: TCloudApiDataSourceConfigSchema, only_basic_settings: bool = True, dbt_parser: Optional[DbtParser] = None
+):
+    def cast_value(value: str, type_: str):
+        if type_ == "integer":
+            return int(value)
+        elif type_ == "boolean":
+            return bool(value)
+        return value
```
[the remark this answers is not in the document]
pik94: Oh, it looks interesting. I know C/C++ has overriding mechanism, and it is quite convenient in contrast to Python.

I see that `from typing import override` fails for now because the corresponding [PEP-698](https://peps.python.org/pep-0698/) will be since Python 3.12 only...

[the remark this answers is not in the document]
nolar: Ah, okay, maybe `from typing_extensions … ` then. Anyway, I have this working since Python 3.7 successfully.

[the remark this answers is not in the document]
pik94: Thanks! I updated.

data_diff/dbt_parser.py
```diff
@@ -0,0 +1,299 @@
+from collections import defaultdict
```
dlawin, line 1: to help me review: are there changes in addition to moving this class to a file? Or is it just moving the code?

pik94: No, there are no changes, just moved the class to a separate file.

PR #471, opened 2023-04-03
Update docs and readme to align with Docs v3

williebsweet:
Update python docs and readme to algin with Docs v3.

Involves moving some documentation back to markdown files in this repo and adjusting links that point to docs.datafold.com.

PR #493, opened 2023-04-11
fix up links in readme

leoebfolsom:
There were several broken and/or irrelevant links since the launch of docs.datafold.com 3.0, as well as an outdated Loom.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ data-diff is a **free, open-source tool** that enables data professionals to det
 
 ## Documentation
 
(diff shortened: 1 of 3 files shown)


dlawin:
> I think there is another open PR for some of this

this? https://github.com/datafold/data-diff/pull/471/files

williebsweet:
@dlawin Yes. 

@leoebfolsom Now that I am looking at it, I think we can split that one in half - the README updates and the Python API doc updates. The README updates can be added to this one

williebsweet (approved):

PR #496, opened 2023-04-12
support insecure_mode key for snowflake profile

### LINKED ISSUE #494: --dbt add support for Snowflake connection that uses insecure mode
I'm unable to connect to Snowflake when running `data-diff --dbt` because we use insecure mode for our dbt<>Snowflake connection and data-diff is still trying to connect using the secure method. As a result, `data-diff --dbt` is giving the error `ERROR - WARNING!!! Using fail-open to connect. Driver is connecting to an HTTPS endpoint without OCSP based Certificate Revocation checking as it could not obtain a valid OCSP Response to use from the CA OCSP responder.`

Support connecting to Snowflake when insecure mode is on, as described here: https://community.snowflake.com/s/article/How-to-turn-off-OCSP-checking-in-Snowflake-client-drivers
In dbt, insecure mode is set in the dbt profile file (profiles.yml) (set via `insecure_mode: true`). When set to true, I'd like data-diff to pull this field and pass it along when connecting to Snowflake so we avoid the above error. Example profiles.yml with insecure mode turned on:
```
---
default:
  target: user
  outputs:
    prod:
      type: snowflake
      account: XXXXXX.us-east-1.privatelink
      user: DATA_BUILD_TOOL_SERVICE
      role: DATA_TRANSFORMER
      insecure_mode: true   <----this is the field that should get pulled and passed through to the Snowflake connection

      private_key_path: "{{ env_var('SNOWFLAKE_KEYPATH_DBTS') }}"
      private_key_passphrase: "{{ env_var('SNOWFLAKE_KEYPASS_DBTS') }}"
```

**Additional context**
Interestingly, in Snowflake we can see that running `data-diff --dbt` successfully kicks off the queries for the diff (and they complete), but the diff results never get spit back out in terminal (terminal instead spits out the error pasted above).

@dlawin asked me to create this issue in this [Slack thread](https://getdbt.slack.com/archives/C03D25A92UU/p1681328794544729?thread_ts=1677252291.169159&cid=C03D25A92UU)

dlawin:
Resolves #494

### DIFF
--- a/data_diff/dbt_parser.py
+++ b/data_diff/dbt_parser.py
@@ -144,6 +144,7 @@ class DbtParser:
                 "warehouse": credentials.get("warehouse"),
                 "role": credentials.get("role"),
                 "schema": credentials.get("schema"),
+                "insecure_mode": credentials.get("insecure_mode") or False,
             }
             self.threads = credentials.get("threads")
             self.requires_upper = True
--- a/data_diff/dbt_parser.py
+++ b/data_diff/dbt_parser.py
@@ -144,7 +144,7 @@ class DbtParser:
                 "warehouse": credentials.get("warehouse"),
                 "role": credentials.get("role"),
                 "schema": credentials.get("schema"),
-                "insecure_mode": credentials.get("insecure_mode") or False,
+                "insecure_mode": credentials.get("insecure_mode", False),
             }
             self.threads = credentials.get("threads")
             self.requires_upper = True


data_diff/dbt_parser.py
```diff
@@ -144,6 +144,7 @@ def set_connection(self):
                 "warehouse": credentials.get("warehouse"),
                 "role": credentials.get("role"),
                 "schema": credentials.get("schema"),
+                "insecure_mode": credentials.get("insecure_mode") or False,
```
pik94, line 147: **nit**: `credentials.get("insecure_mode", False)

pik94 (approved):

data_diff/dbt_parser.py
dlawin: ty simplified this

Issue #444, opened 2023-03-16
--dbt --select


dlawin:
**Is your feature request related to a problem? Please describe.**
I don't always want to run diffs against my last dbt run. I'd like to choose what to diff

**Describe the solution you'd like**
Support selecting models using dbt syntax

e.g. `data-diff --dbt --select model+`
- runs diff against model and any downstream models

dave-connors-3:
@dlawin I am happy to take a swing at this!

dlawin:
> @dlawin I am happy to take a swing at this!

that would be awesome!

This method is where we are getting a list of models to diff currently https://github.com/datafold/data-diff/blob/9cb0e1dd2c4757f626c708da50083bb297add718/data_diff/dbt.py#L413

I'll let you solution but my initial thought is we can make that more generic, either getting models from the run_results (current state) or the `--select` results

PR #511, opened 2023-04-19
Allow dbt selectors

dave-connors-3:
closes: #444 

Here's some early thinking on how we might support dbt selection syntax as an arg to data-diff! 

In dbt 1.5+ there's a new interface to call out to dbt -- using a `dbt ls` command under the hood, we can take the argument passed to the new flag `--select` or `-s` and return the models that match that selector. 

This is absolutely going to fail some tests, so would love a little bit of guidance on which tests to update, and how they might need to change!


dlawin:
> This is absolutely going to fail some tests, so would love a little bit of guidance on which tests to update, and how they might need to change!

I went ahead and handled the test changes, outside of that I mostly made some stylistic changes and handled a couple exception scenarios.

Really appreciate the contribution @dave-connors-3 , I'm going to tag some team members to review, but I think this is in good shape!

data_diff/dbt.py
```diff
@@ -18,17 +18,6 @@
 logger = getLogger(__name__)
 
 
-def import_dbt():
```
dlawin, line 21: I thought the deletion of this was in master but apparently not. Not needed -- duplicate code

data_diff/dbt_parser.py
```diff
@@ -12,23 +13,34 @@
...
+        from dbt.cli.main import dbtRunner
+    except ImportError:
+        dbtRunner = None
+
+    if dbtRunner is not None:
+        dbt_runner = dbtRunner()
+    else:
+        dbt_runner = None
+
+    return parse_run_results, parse_manifest, ProfileRenderer, yaml, dbt_runner
```
nolar, line 35: @dlawin Is it worth aligning these changes with the similar ones in #544? — There, these modules are not optional anymore, and the importing function is gone, the imports are moved to the module level.

data_diff/dbt_parser.py
```diff
@@ -68,7 +86,60 @@ def get_datadiff_variables(self) -> dict:
         vars = get_from_dict_with_raise(self.project_dict, "vars", error_message)
         return get_from_dict_with_raise(vars, "data_diff", error_message)
 
-    def get_models(self):
+    def get_models(self, dbt_selection: Optional[str] = None):
+        dbt_version = parse_version(self.dbt_version)
+        if dbt_selection:
+            if dbt_version.major == 1 and dbt_version.minor >= 5:
```
nolar, line 92: **nit:** Despite this works now, it would break with v2.x, when we forget about such version restrictions. A somewhat better and common approach is lexographical comparison of tuples:

```suggestion
            if (dbt_version.major, dbt_version.minor) >= (1, 5):
```

nolar (approved):

data_diff/dbt_parser.py
dlawin: I am leaning towards merging this PR first and I'll handle in #544

dlawin (approved):

Issue #518, opened 2023-04-20
Specify table filters and include/exclude columns


dlawin:
I don't always want or need to diff every row + column in a model -- also, with larger tables I would like some way to increase the performance of `data-diff --dbt`.

Add support for:
- Table filters https://docs.datafold.com/guides/dbt_advanced_configs#filter-tables
- Specifying columns https://docs.datafold.com/guides/dbt_advanced_configs#filter-tables

This will also increase parity between open source data-diff and Datafold's cloud product.

PR #522, opened 2023-04-20
support datadiff meta filter

dlawin:
Adds support for the meta filter described here:
https://docs.datafold.com/guides/dbt_advanced_configs#filter-tables

The filter expression is applied to both sides of the diff

Resolves partially: #518

### DIFF
--- a/data_diff/cloud/datafold_api.py
+++ b/data_diff/cloud/datafold_api.py
@@ -101,6 +101,8 @@ class TCloudApiDataDiff(pydantic.BaseModel):
     table1: List[str]
     table2: List[str]
     pk_columns: List[str]
+    filter1: Optional[str] = None
+    filter2: Optional[str] = None
 
 
 class TSummaryResultPrimaryKeyStats(pydantic.BaseModel):
--- a/data_diff/dbt.py
+++ b/data_diff/dbt.py
@@ -73,6 +73,7 @@ class DiffVars:
     primary_keys: List[str]
     connection: Dict[str, str]
     threads: Optional[int]
+    where_filter: Optional[str] = None
 
 
 def dbt_diff(
@@ -191,7 +192,16 @@ def _get_diff_vars(
         dev_qualified_list = [dev_database, dev_schema, model.alias]
         prod_qualified_list = [prod_database, prod_schema, model.alias]
 
-    return DiffVars(dev_qualified_list, prod_qualified_list, primary_keys, dbt_parser.connection, dbt_parser.threads)
+    where_filter = None
+    if model.meta:
+        try:
+            where_filter = model.meta["datafold"]["datadiff"]["filter"]
+        except KeyError:
+            pass
+
+    return DiffVars(
+        dev_qualified_list, prod_qualified_list, primary_keys, dbt_parser.connection, dbt_parser.threads, where_filter
+    )
 
 
 def _local_diff(diff_vars: DiffVars) -> None:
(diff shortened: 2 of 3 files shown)


data_diff/dbt.py
```diff
@@ -191,7 +192,16 @@ def _get_diff_vars(
         dev_qualified_list = [dev_database, dev_schema, model.alias]
         prod_qualified_list = [prod_database, prod_schema, model.alias]
 
-    return DiffVars(dev_qualified_list, prod_qualified_list, primary_keys, dbt_parser.connection, dbt_parser.threads)
+    where = None
+    if model.meta:
+        try:
+            where = model.meta["datafold"]["datadiff"]["filter"]
```
dlawin, line 198: very open to suggestion here -- not very pretty

data_diff/cloud/datafold_api.py
```diff
@@ -101,6 +101,8 @@ class TCloudApiDataDiff(pydantic.BaseModel):
     table1: List[str]
     table2: List[str]
     pk_columns: List[str]
+    filter1: Optional[str]
```
pik94, line 104:
```suggestion
    filter1: Optional[str] = None
    filter2: Optional[str] = None
```

data_diff/dbt.py
```diff
@@ -73,6 +73,7 @@ class DiffVars:
     primary_keys: List[str]
     connection: Dict[str, str]
     threads: Optional[int]
+    where: Optional[str]
```
pik94, line 76:
```suggestion
    where: Optional[str] = None
```

data_diff/dbt.py
```diff
@@ -191,7 +192,16 @@ def _get_diff_vars(
         dev_qualified_list = [dev_database, dev_schema, model.alias]
         prod_qualified_list = [prod_database, prod_schema, model.alias]
 
-    return DiffVars(dev_qualified_list, prod_qualified_list, primary_keys, dbt_parser.connection, dbt_parser.threads)
+    where = None
```
pik94, line 195: Maybe `where_filter` or `where_clause` are better because more readable.

data_diff/dbt.py
pik94: **nit**: For this or future PRs.

A datafold datadiff section might contain other useful different parameters such as include/exclude_columns, time_travel columns etc which we may want to use in the future for running cloud diffs.

I think it is worth to have a special classmethod of `DbtParser` or separate function to extract from a model that section if it exists.
```
           # dbt_parser.py
           class TDatafoldDatadiffModelParams(pydantic.BasicModel):
               filter_: Optional[str] = None
               include_columns: Optional[List[str]] = None
               exclude_columns: Optional[List[str]] = None

            class DbtParser:  
                  @classmethod
                  def get_datafold_datadiff_params(cls, model) -> TDatafoldDatadiffModelParams:
                      # implementation
            
            # dbt.py     
            datafold_datadiff_params = dbt_parser.get_datafold_datadiff_params()
            return DiffVars(
                 dev_qualified_list, prod_qualified_list, primary_keys, dbt_parser.connection, dbt_parser.threads, datafold_datadiff_params.filter_)
```

data_diff/dbt.py
dlawin: done

data_diff/dbt.py
dlawin: good point. I'm about to add include/exclude columns and I will refactor in that PR

vvkh (approved):

PR #531, opened 2023-04-26
add support for two part id

### LINKED ISSUE #530: Add support for empty database variable in Databricks
**Describe the bug**
See #448

dlawin:
Resolves #530

This will also make the `prod_database:` field in dbt_project.yml optional. When that is not included, use the database from the local manifest.json (prod_db == dev_db)

This can be `None` for some dbt adapters like Databricks, so `None` values are filtered out.

### DIFF
--- a/data_diff/dbt.py
+++ b/data_diff/dbt.py
@@ -161,8 +161,8 @@ def _get_diff_vars(
         prod_qualified_list = [x.upper() for x in [prod_database, prod_schema, model.alias] if x]
         primary_keys = [x.upper() for x in primary_keys]
     else:
-        dev_qualified_list = list(filter(None, [dev_database, dev_schema, model.alias]))
-        prod_qualified_list = list(filter(None, [prod_database, prod_schema, model.alias]))
+        dev_qualified_list = [x for x in [dev_database, dev_schema, model.alias] if x]
+        prod_qualified_list = [x for x in [prod_database, prod_schema, model.alias] if x]
 
     where_filter = None
     if model.meta:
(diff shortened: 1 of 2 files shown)


data_diff/dbt.py
```diff
@@ -162,12 +157,12 @@ def _get_diff_vars(
...
-        dev_qualified_list = [x.upper() for x in [dev_database, dev_schema, model.alias]]
-        prod_qualified_list = [x.upper() for x in [prod_database, prod_schema, model.alias]]
+        dev_qualified_list = [x.upper() for x in [dev_database, dev_schema, model.alias] if x]
+        prod_qualified_list = [x.upper() for x in [prod_database, prod_schema, model.alias] if x]
         primary_keys = [x.upper() for x in primary_keys]
     else:
-        dev_qualified_list = [dev_database, dev_schema, model.alias]
-        prod_qualified_list = [prod_database, prod_schema, model.alias]
+        dev_qualified_list = list(filter(None, [dev_database, dev_schema, model.alias]))
+        prod_qualified_list = list(filter(None, [prod_database, prod_schema, model.alias]))
```
nolar, line 165: Generally, list comprehensions are considered to be better than iterators, especially than iterators instantly converted to lists. So, to align with the style a few lines above:

```suggestion
        dev_qualified_list = [x for x in [dev_database, dev_schema, model.alias] if x]
        prod_qualified_list = [x for x in [prod_database, prod_schema, model.alias] if x]
```

nolar (approved):

PR #535, opened 2023-04-28
support include/exclude meta config

### LINKED ISSUE #518: Specify table filters and include/exclude columns

dlawin:
Adds `--dbt` support for the advanced config described here:
https://docs.datafold.com/guides/dbt_advanced_configs#include--exclude-columns

Adds a new type `TDatadiffModelConfig` per @pik94 feedback here https://github.com/datafold/data-diff/pull/522#discussion_r1173506644

Reworks DiffVars type -> TDiffVars + pydantic to increase consistency

Resolves #518

### DIFF
--- a/tests/test_dbt.py
+++ b/tests/test_dbt.py
@@ -73,7 +73,7 @@ class TestDbtParser(unittest.TestCase):
         with self.assertRaises(Exception):
             _ = DbtParser.get_models(mock_self, selection)
         mock_self.get_dbt_selection_models.assert_not_called()
-        
+
     def test_get_models_no_runner(self):
         mock_self = Mock()
         mock_self.project_dir = Path()


data_diff/dbt.py
```diff
@@ -202,18 +205,25 @@ def _local_diff(diff_vars: DiffVars) -> None:
         rich.print(diff_output_str)
         return
 
-    mutual_set = set(table1_columns) & set(table2_columns)
-    table1_set_diff = list(set(table1_columns) - set(table2_columns))
-    table2_set_diff = list(set(table2_columns) - set(table1_columns))
+    column_set = set(table1_columns).intersection(table2_columns)
+    table1_diff = set(table1_columns).difference(table2_columns)
+    table2_diff = set(table2_columns).difference(table1_columns)
```
nolar, line 210: Out of curiosity, why? The original version with `&` & `-` seemed more readable.

data_diff/dbt_parser.py
```diff
@@ -140,7 +162,7 @@ def get_connection_creds(self) -> Tuple[Dict[str, str], str]:
 
         return credentials, conn_type
 
-    def set_connection(self):
+    def set_connection(self) -> dict:
```
nolar, line 165: This method has no single `return` in it, so it returns `None`, doesn't it?

nolar (approved):

data_diff/dbt_parser.py
dlawin: ty for catching this, fixed

PR #537, opened 2023-04-28
Clarify dependency error

### LINKED ISSUE #525: "could not import 'dbt' package" error is confusing
The exception thrown [here](https://github.com/datafold/data-diff/blob/2f3e83da0ceb5fbec5e8acf7dbf1aa90160ae7f7/data_diff/dbt.py#L27) frequently causes confusion as it is more commonly thrown when `dbt_artifacts_parser` is missing.

The guidance to resolve the issue is correct however.

Make the error more generic, something like `Please run pip install 'data-diff[dbt]' to install missing package dependencies.`

dlawin:
Make the error more clear when dependencies are missing 

Removes duplicated method

Resolves #525

### DIFF
--- a/data_diff/dbt.py
+++ b/data_diff/dbt.py
@@ -18,17 +18,6 @@ from .dbt_parser import DbtParser, PROJECT_FILE
 logger = getLogger(__name__)
 
 
-def import_dbt():
-    try:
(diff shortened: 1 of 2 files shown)


dlawin:
Resolved in #544

PR #544, opened 2023-05-03
remove optional flag from --dbt dependencies

### LINKED ISSUE #525: "could not import 'dbt' package" error is confusing

### LINKED ISSUE #526: Several dependencies being "Extras" makes installation complex
There are several ways to install data-diff depending on the usage and database provider.

Examples:
`--dbt` user on snowflake: `pip install data-diff 'data-diff[snowflake,dbt]'`
`--dbt` user on bigquery: `pip install data-diff 'data-diff[dbt]' google-cloud-bigquery`
non `--dbt` user on postgres: `pip install data-diff 'data-diff[postgres]'`
... etc.

Reduce the number of pip "extras" to simplify installation

Balancing the tradeoff of increasing package size

**Additional context**
Installation instructions: https://docs.datafold.com/development_testing/open_source

dlawin:
Resolves #526 
Resolves #525 

Makes the following packages non-optional:
```
dbt-artifacts-parser = {version="^0.2.5"}
dbt-core = {version="^1.0.0"}
```

The intent is to simplify installation for `--dbt` users.

We can tell folks to `pip install data-diff`, and they will be prompted with a single instruction for their database provider vs. the current two. Something like:
Screenshot 2023-05-03 at 2 23 51 PM

This will increase the number of required dependencies


dlawin:
this would essentially revert https://github.com/datafold/data-diff/issues/412

leoebfolsom:
BQ & Databricks?

dlawin:
> BQ & Databricks?

We could add those as well, the only reason I excluded is that they are not listed as dependencies at all currently (not even as optional ones)

leoebfolsom:
> We could add those as well, the only reason I excluded is that they are not listed as dependencies at all currently (not even as optional ones)

Got it--I just want to make sure users of BQ and Databricks are able to get started with the same ease as the other DBs you listed, and they don't have to run these commands (which are currently listed in [the documentation](https://docs.datafold.com/development_testing/open_source/)):

- `pip install data-diff 'data-diff[dbt]' google-cloud-bigquery`
- `pip install data-diff 'data-diff[databricks,dbt]' -U`

data_diff/sqeleton/databases/base.py
```diff
@@ -59,10 +59,10 @@ def _inner():
             try:
                 return f()
             except ModuleNotFoundError as e:
-                s = text
                 if package:
-                    s += f"You can install it using 'pip install data_diff[{package}]'."
-                raise ModuleNotFoundError(f"{e}\n\n{s}\n")
+                    print(f"Missing database dependency, complete setup by running 'pip install data_diff[{package}]'.\n")
```
leoebfolsom, line 63: @dlawin I would simply change this to:

```
Please complete the setup by running 'pip install data_diff[{package}]'.
```

data_diff/dbt_parser.py
```diff
@@ -49,7 +40,10 @@ def legacy_profiles_dir() -> Path:
 
 class DbtParser:
     def __init__(self, profiles_dir_override: str, project_dir_override: str) -> None:
-        self.parse_run_results, self.parse_manifest, self.ProfileRenderer, self.yaml = import_dbt()
+        self.parse_run_results = parse_run_results
+        self.parse_manifest = parse_manifest
+        self.ProfileRenderer = ProfileRenderer
+        self.yaml = yaml
```
nolar, line 46: Maybe we should remove this `self.whatever` approach here and use the functions/classes directly as they are imported (as we do everywhere all the time normally) — to reduce the unnecessary code complexity and contributor confusion.

data_diff/sqeleton/databases/base.py
```diff
@@ -59,10 +59,10 @@ def _inner():
...
+                    exit()
+                raise ModuleNotFoundError(f"{e}\n")
```
nolar, line 65: Since the extra hint is gone and it is now the same exception of the same type with the same message, we should rather do `raise` (no arguments) in the end and let the original exception with the original stack trace escalate.

data_diff/dbt_parser.py
dlawin: good point, updated

data_diff/sqeleton/databases/base.py
dlawin: The issue I see is around presenting an expected user workflow as an error

data_diff/sqeleton/databases/base.py
dlawin: fixed

data_diff/sqeleton/databases/base.py
dlawin: I reverted to the original raise -- showing the word "error" is pretty minor

leoebfolsom (approved):

PR #547, opened 2023-05-03
enhance cloud event metadata

dlawin:
Adds additional fields to diff end events in order to differentiate between non-unique diff_ids (diff ids from self-hosted datafold)


kylemcnair (approved):
looks good from the perspective of the data being tracked

PR #555, opened 2023-05-05
instantiate MACRO_DEBUGGING global for dbt

dlawin:
Without setting this flag, we're seeing an exception in dbt 1.5

Assuming it will not hurt earlier versions to set it

```
  File "/Users/dan/Library/Caches/pypoetry/virtualenvs/data-diff-gej9eYRL-py3.10/lib/python3.10/site-packages/dbt/clients/jinja.py", line 102, in _compile
    macro_debugging = get_flags().MACRO_DEBUGGING
AttributeError: 'Namespace' object has no attribute 'MACRO_DEBUGGING'
```


dlawin:
I tested previous dbt versions

PR #558, opened 2023-05-06
missing --dbt  packages

dlawin:
Add these packages explicitly to avoid ModuleNotFound errors when using `data-diff --dbt` before installing the [dbt] extras


dlawin:
https://github.com/datafold/data-diff/pull/544 will resolve this as well

PR #579, opened 2023-05-12
Use proper casing policy for --cloud diffs

vvkh:
Datafold API attempts to infer the correct casing based on the presence of tables in the schema. However, to avoid unnecessary SQL queries to the user database, we should provide the correct casing whenever possible. Currently we do that only for client-side diffs but not for `--cloud` diffs.


dlawin (approved):
ty @vvkh

PR #589, opened 2023-05-22
Add custom exceptions for raises in dbt_parser

dlawin:
Changes this module to not use generic exception types so error events can be more understandable.

Also involved changing utils.get_dict_with_raise to accept an exception parameter

### DIFF
--- a/data_diff/dbt_parser.py
+++ b/data_diff/dbt_parser.py
@@ -12,18 +12,18 @@ from dbt_artifacts_parser.parser import parse_run_results, parse_manifest
 from dbt.config.renderer import ProfileRenderer
 
 from data_diff.errors import (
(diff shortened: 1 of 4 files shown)


data_diff/dbt_parser.py
```diff
@@ -11,6 +11,21 @@
 from dbt_artifacts_parser.parser import parse_run_results, parse_manifest
 from dbt.config.renderer import ProfileRenderer
 
+from data_diff.errors import (
```
pik94, line 14: I would recommend to add a prefix `DataDiff` to that error names because when errors happen it will help to understand that they occur in data diff, not in a dbt package.

dlawin: updated

PR #605, opened 2023-06-07
--dbt option to sign up for email notifications

dlawin:
Prompts the user once on if they'd like to sign up for email notifications:
Screenshot 2023-06-07 at 3 52 22 PM

Stores if the user has already been asked in .datadiff.toml by adding a value `asked_for_email = ""`

### DIFF
--- a/data_diff/dbt.py
+++ b/data_diff/dbt.py
@@ -1,4 +1,5 @@
 import os
+import re
 import time
 import webbrowser
 from typing import List, Optional, Dict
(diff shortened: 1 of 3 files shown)


data_diff/dbt.py
```diff
@@ -389,3 +389,23 @@ def _cloud_diff(diff_vars: TDiffVars, datasource_id: int, api: DatafoldAPI, org_
...
+
+
+def _email_signup() -> None:
+    if bool_ask_for_email():
+        email_input = Prompt.ask(
+            "\nWould you like to be notified when a new data-diff version is available?\n\nEnter email or leave blank to opt out (we'll only ask once).\n",
+            default="",
+            show_default=False,
+        )
+        email = email_input.strip()
```
sklinov, line 408: Hey, @dlawin! Is it possible to add some validation here?
We have some handy e-mail RegExp at our disposal

dlawin: I was using this at one point https://pypi.org/project/email-validator/

Do you think it's worth it if we're going to need to validate on the backend anyways?

dlawin: Thinking about it a bit I don't see a downside if we validate twice. Let me know how this looks now

sklinov (approved):

data_diff/dbt.py
sklinov: @dlawin As long as we're using the same regex for both validations - we're good

PR #613, opened 2023-06-15
remove prompt to create API key, send user straight to docs

leoebfolsom:
Currently, if the API key is not found when the `--cloud` flag is used, the user is guided to generate an API key, go through data source creation, etc.

We should instead simply provide them with a link to the instructions for storing an API key and adding the data source id to `dbt_project.yml`: https://docs.datafold.com/development_testing/cloud


leoebfolsom:
Let's see if tests pass. @dlawin I will not at all be offended if this PR is closed and you takes a different approach to addressing the ticket, but I wanted to take a stab at it.

leoebfolsom:
@dlawin I think I may have resolved the merge conflicts 🤞

dlawin (approved):

dlawin (approved):

PR #614, opened 2023-06-16
log improvements and display diff status

dlawin:
Improves the `--debug` log readability by using [RichHandler](https://rich.readthedocs.io/en/stable/logging.html). This format is also used without the debug flag for warnings and errors

Screenshot 2023-06-16 at 10 10 28 AM

Adds a new custom logging handler LogStatusHandler. This is used to display diff status and some detail for local and cloud `--dbt` diffs.

Local dbt diff:
Screenshot 2023-06-16 at 11 38 14 AM

Cloud dbt diff:
Started (it's not ideal doubling some of the info here, but it seems a rich.status.Status can't display a clickable link... otherwise one line would be great per model)
Screenshot 2023-06-16 at 10 08 13 AM
One completed
Screenshot 2023-06-15 at 4 26 07 PM

### DIFF
--- a/data_diff/__main__.py
+++ b/data_diff/__main__.py
@@ -6,16 +6,17 @@ import time
 import json
 import logging
 from itertools import islice
-from typing import Optional
+from typing import Dict, Optional
 
 import rich
+from rich.logging import RichHandler
 import click
 
 from data_diff.sqeleton.schema import create_schema
 from data_diff.sqeleton.queries.api import current_timestamp
 
 from .dbt import dbt_diff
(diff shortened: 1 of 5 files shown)


data_diff/__main__.py
```diff
@@ -38,6 +36,21 @@
 set_entrypoint_name("CLI")
 
 
+def _get_log_handlers() -> Tuple[RichHandler, LogStatusHandler]:
```
dlawin, line 39: this could use a better location than __main__

PR #638, opened 2023-07-09
Add column types to json output

vvkh:
Add info about column types to `--json` output. The output includes both raw type (`type`) and logical type (`kind`).
No breaking changes to the `--json` format, only new fields were added.


sklinov:
@vvkh @dlawin, I suggest we keep the same naming as we have in the Cloud for consistency.
So the native type from the database should be `db_type,` and the mapped type would be `type`.

dlawin (approved):

PR #643, opened 2023-07-17
add --dbt support for --columns

dlawin:
--columns are used as the TDiffVars.include_columns when present


sklinov:
@dlawin Would this allow diffing only selected columns?

dlawin:
> @dlawin Would this allow diffing only selected columns?

@sklinov yes exactly, this is analogous to using include_columns in dbt meta

PR #644, opened 2023-07-17
poll more frequently when using --cloud

dlawin:
Poll more frequently when using `--dbt --cloud`

Biggest change is that the backoff is now additive, it just increases by 1 sec each iteration to a max of 20 sec vs. the former doubling.

Also created custom exceptions for existing generic exceptions since I was in there.


leoebfolsom (approved):

sklinov (approved):

leoebfolsom:
@dlawin I'll let you merge so you can manage it with respect to deployment (I don't want to "merge and run")

Issue #645, opened 2023-07-18
Profiles dir overwritten in the code for DBT version < 1.3.0


sebaap:
**Describe the bug**
The profiles dir is being overwritten in the code ([this lines](https://github.com/datafold/data-diff/blob/eb22d3e9a8b1a098f113c7a37dec9a8efb25c025/data_diff/dbt_parser.py#L214-L215)) after passed as an argument.
If I execute:
`data_diff --dbt --dbt-profiles-dir my/custom/dir`
Then I get
`ERROR    No profile 'my_dbt_profile' found in '/Users/myuser/.dbt/profiles.yml'`

I'm guessing this is not the intended behaviour as I believe the argument should take precedence in this case.

**Describe the environment**

I'm using the master branch of this repo ([v0.7.14](https://github.com/datafold/data-diff/commit/eb22d3e9a8b1a098f113c7a37dec9a8efb25c025)) and DBT v1.2.6 but i guess any < 1.3.0 will have this issue

dlawin:
> happy to submit a PR if someone confirms this is a bug and help is needed btw

Good catch @sebaap, this is not intended. The \~intent\~ is to follow the default profiles.yml location based on dbt version, but for the flag to take precedence. 

```
        if dbt_version < parse_version("1.3.0"):
            self.profiles_dir = legacy_profiles_dir()
```

I think it makes sense to pull this out of the method and into the class \_\_init\_\_ and conditionally set it there based on dbt_version + the presence of `profiles_dir_override`

The dbt version is available [there](https://github.com/datafold/data-diff/blob/eb22d3e9a8b1a098f113c7a37dec9a8efb25c025/data_diff/dbt_parser.py#L116C10-L116C25)


We'd appreciate the contribution if you have time! Assigning this to you but lmk if you need to hand it off

dlawin:
One other thing to keep in mind is this [env_var](https://github.com/datafold/data-diff/blob/eb22d3e9a8b1a098f113c7a37dec9a8efb25c025/data_diff/__main__.py#L241)

PR #651, opened 2023-07-20
fix profile_dir argument overwritten by legacy dir for DBT < 1.3.0

sebaap:
Resolves https://github.com/datafold/data-diff/issues/645

This PR fixes an issue with the `profiles_dir` provided in the cli arguments being overwritten for DBT versions < 1.3.0, even though it should take precedence.
This will also cover the case when the `DBT_PROFILES_DIR` env var is set, as that's handled by click and should come in the `profiles_dir_override` argument.


dlawin (approved):
thanks for the contribution! @sebaap

Issue #646, opened 2023-07-18
Get prod alias from `--state` manifest file when provided


sebaap:
**Is your feature request related to a problem? Please describe.**
We have an uncommon BigQuery setup where our prod environment consist of multiple GCP projects, but dev is a single project. In this setup the name of the underneath tables for each model is different between the two environments, so currently data-diff doesn't work for us because it tries to use the same model alias in both environments.

**Describe the solution you'd like**
When providing a prod manifest via the `--state`, take the alias for prod from the manifest instead of using the same one from the `run_results` file. This should be fairly easy to implement and would allow us to use data-diff :)

**Describe alternatives you've considered**
With my very limited familiarity with the code, I don't see any other alternative to be honest.
Migrating to a single project setup is not an option for us as it would require a huge effort.

**Additional context**
I'm happy to submit a PR if help is needed, as I have a working version of this already.

dlawin:
I think this was overlooked, no reason not to use the prod alias when we have the manifest here https://github.com/datafold/data-diff/blob/eb22d3e9a8b1a098f113c7a37dec9a8efb25c025/data_diff/dbt.py#L171

dlawin:
> I'm happy to submit a PR if help is needed, as I have a working version of this already.

Please do and don't worry about the tests if some break, happy to take it the last mile

PR #652, opened 2023-07-20
get prod alias from manifest file when provided

sebaap:
Resolves https://github.com/datafold/data-diff/issues/646

Currently the model alias is extracted from the run_results file even when the manifest file is provided, which doesn't allow to use data-diff when the prod aliases are different.
This PR modifies this behaviour and extracts the prod aliases from the manifest file when provided.


dlawin (approved):
thanks for the contribution @sebaap !

PR #654, opened 2023-07-24
add TRIGGERED_BY env variable support

vvkh:
Add support for the `TRIGGERED_BY` env variable that overrides source of Rudderstack events

### DIFF
--- a/data_diff/__main__.py
+++ b/data_diff/__main__.py
@@ -33,7 +33,7 @@ COLOR_SCHEME = {
     "-": "red",
 }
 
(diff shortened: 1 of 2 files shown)


sklinov (approved):

data_diff/__main__.py
```diff
@@ -33,7 +33,7 @@
     "-": "red",
 }
 
-set_entrypoint_name("CLI")
+set_entrypoint_name(os.getenv("TRIGGERED_BY", "CLI"))
```
dlawin, line 36: nit but I feel like this should be more specific. e.g. `DATAFOLD_TRIGGERED_BY`

PR #662, opened 2023-08-01
Fix for more than 50 fields in Postgres

kindly:
Postgres does not allow functions that have more that have more than 100 arguments. When using the concat function, this limits comparisons to less than 50 fields.
Using `||` for concat like the Oracle variant fixes this.

### DIFF
--- a/data_diff/sqeleton/databases/postgresql.py
+++ b/data_diff/sqeleton/databases/postgresql.py
@@ -1,3 +1,4 @@
+from typing import List
 from ..abcs.database_types import (
     DbPath,
     JSON,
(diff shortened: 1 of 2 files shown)


kindly:
Is there anything I can do to help get this merged.  As I am trying to do a comparison of over 50 fields in postgres and without this patch, it crashes, so I am having to use my own fork.

dlawin:
Relates to https://github.com/datafold/data-diff/issues/397

dlawin:
> Is there anything I can do to help get this merged. As I am trying to do a comparison of over 50 fields in postgres and without this patch, it crashes, so I am having to use my own fork.

Thanks for opening this @kindly 

If you're up for it adding a test case [here](https://github.com/datafold/data-diff/blob/master/tests/test_postgresql.py) would speed things along

There's dev setup info [here](https://github.com/datafold/data-diff/blob/master/CONTRIBUTING.md)

kindly:
@dlawin 

I have added a test which will fail if you remove the new code.

I had difficulty reproducing the issue in the tests, and only by using the `uuid` like the other test could I make it fail.  This is because the error caused by the checksumming/segmenting, and I think that is only triggered in some conditions.

dlawin:
> I have added a test which will fail if you remove the new code.
> 
> I had difficulty reproducing the issue in the tests, and only by using the `uuid` like the other test could I make it fail. This is because the error caused by the checksumming/segmenting, and I think that is only triggered in some conditions.

appreciate it @kindly I will take a closer look early next week and get it in a release!

nolar (approved):

PR #671, opened 2023-08-07
Beautify readme

sungchun12:
The readme is the first impression for data-diff. It needs to show tasteful pizazz! 

- added beautiful images
- cleaned up the content flow


README.md
```diff
@@ -1,25 +1,61 @@
...
+1. Make a change to some SQL code
+2. Run the SQL code to create a new dataset
+3. Compare the dataset with its production version or another iteration
+
+  <p align="left">
+  <img alt="dbt" src="https://seeklogo.com/images/D/dbt-logo-E4B0ED72A2-seeklogo.com.png" width="10%" />
+  </p>
+  
+<details>
+<summary> Integrates with dbt Core and dbt Cloud to seamlessly compare local development to production datasets. </summary>
```
leoebfolsom, line 32: This is something that I was going to update, and already discussed with Gleb needs to be updated (he agreed):  

`data-diff` does not integrate with dbt Cloud. I would say that _Datafold_ integrates with dbt Cloud, but it's misleading to say that `data-diff` integrates with dbt Cloud. Because in order to use it with dbt Cloud, well, you can't -- you have to install dbt Core.

This came up recently because I got a question about the statement ("how does data-diff integrate with dbt Cloud?") pointing to this line in the readme. I had to say, "you're right--it doesn't. I'll update the readme.

leoebfolsom: Here's a suggested update:

> data-diff integrates with dbt Core to seamlessly compare local development to production datasets. dbt Cloud users should check out [Datafold's out-of-the-box development testing integration with dbt Cloud](https://www.datafold.com/data-deployment-testing).

leoebfolsom (approved):
Approved with one comment and requested edit ✨

dlawin (approved):

docs/data-diff-logo.png
glebmezh, line 1: TBH I find this image confusing.
1. the SQL-formatted comment "Regression Testing Project"
2. The output follows "pip install data-diff" (the install command) which is confusing, if we want to show the output we should give an example of an actual `data-diff` command.
3. The visualization itself has something going on on the left side and then the diff on the right and kind of overcomplicates the "compare datasets" action.

leoebfolsom: Agree with these critiques -- @sungchun12  maybe we could merge without the image and then iterate on the image? Not strongly opinionated.

sungchun12: If the image is confusing there, then worth noting that we use the same image for the grand home page here: https://www.datafold.com/open-source-data-diff

Happy to swap it out for another image. I'm not emotionally attached to this one in particular. An image helps captivate and entices the reader to keep scrolling down. 

What do you all think about this?
62a30ae50a5023399465c020_OS-Datadiff-feature-chain

sungchun12: @glebmezh One more review please!

leoebfolsom: I like that new image, @sungchun12 -- defer to Gleb

glebmezh: let's do it!

Issue #669, opened 2023-08-04
Add support for python-oracledb


kangshung:
A new driver for Oracle, a successor to cx_Oracle: https://python-oracledb.readthedocs.io/en/latest/index.html

PyPI: https://pypi.org/project/oracledb/

PR #681, opened 2023-08-15
Switch from `cx_Oracle` to `oracledb`

nolar:
`cx_Oracle` is officially deprecated since 2021 (2 years go) and no new released were made. The new official library is `oracledb`. This warning is the first sentence in cx_Oracle's docs.

New projects, such as data-diff, should use the new library. Luckily, `oracledb` is mostly a drop-in replacement for `cx_Oracle`, so no code changes are needed.

Besides, `oracledb` provides 2 modes: the "thin" mode and the "thick" mode. The latter (`"thick") one requires binary libraries, the former one ("thin") works in pure Python. This simplifies the Oracle usage for users.


sungchun12 (commented):
So be default this driver is using "thin" mode. Is there ever a reason to use "thick" mode when data diffing?

sungchun12:
Quick chatgpt reference. May be worth adding thick mode into Datafold for xdb data diffing if thick mode has a significant performance boost.

dlawin (approved):

dlawin:
Resolves #669

choudrybabu:
Hello @dlawin 
is there a way to force Thick mode ?
as our oracle db pwd are case insensitive thin mode isnt working , can you please let me know if there is any alternative approach 
i have tried my luck with our DBA team to make it case sensitive , but no luck

PR #688, opened 2023-08-25
Refactor Artifacts Parser to be Native so it's less brittle with each dbt version change

sungchun12:
**Problem**: Each time a new dbt version is released, it results in running fire drills to fix breaking changes to the dbt integration specifically because of version upgrades to `run_results.json` and `manifest.json`. This is pronounced because `data-diff` relies on a solo-developer maintained package to parse the json docs into python objects. https://github.com/yu-iskw/dbt-artifacts-parser

**Solution**: Build native json/dictionary parsing within `data-diff` and focus on stable fields so that new version upgrades are immaterial to how `data-diff` interacts with dbt.

**Technical Approach**
I create native pydantic schema enforcers for both the `manifest.json` and `run_results.json` with only the fields necessary for `data-diff` to work properly. The fields `data-diff` reads in are relatively stable across artifact schema changes and can handle malformed artifact schemas if they're manipulated by a process outside of dbt. This makes this piece of the codebase lighter weight as the `dbt-artifacts-parser` enforces more artifact schema validation than needed. When future bugs arise because of breaking changes to the artifact schemas, we can simply adjust the native schema validator in `data-diff` vs. relying on the solo-developer maintained package. 

**Testing Approach**
I installed multiple versions of `dbt-snowflake` and `dbt-core` to produce `manifest.json` and `run_results.json` artifacts for version `>=1.0.0`. dbt versions `<1.0.0` are no longer supported by dbt Labs, so they are not tested. Testing project: https://github.com/datafold/datafold-demo-sung

Note: Only one version of `run_results.json` is maintained by dbt Labs which is v4.


data_diff/dbt_parser.py
```diff
@@ -81,19 +81,84 @@ def legacy_profiles_dir() -> Path:
     return Path.home() / ".dbt"
 
 
-class TDatadiffModelConfig(pydantic.BaseModel):
+class TDatadiffModelConfig(BaseModel):
```
dlawin, line 84: I think we could benefit by moving our pydantic models to a separate module -- like dbt_models.py or something. Non-blocking

dlawin:
Approach looks great, this will simplify things in an elegant way, appreciate it @sungchun12

data_diff/dbt_parser.py
sungchun12: Updated to a separate module! Good idea as there's enough to justify it.

sungchun12: brought it back because I'd have to make too many out of scope changes across the codebase

dlawin (approved):
This is great, just need to remove the submodule before we merge

tests/test_dbt_config_validators.py
```diff
@@ -0,0 +1,34 @@
...
+
+class TestManifestJsonConfig(unittest.TestCase):
+    def test_manifest(self):
+        # https://docs.getdbt.com/reference/artifacts/manifest-json
+        versions = ["v4", "v5", "v6", "v7", "v8", "v9", "v10", "v11"]
+
+        for version in versions:
+            with self.subTest(version=version):
+                with open(Path(MANIFEST_PATH, f"manifest_{version}.json"), "r", encoding="utf-8") as manifest:
+                    ManifestJsonConfig.parse_obj(json.load(manifest))
```
dlawin, line 34: non blocking: if we wanted to be very careful we could assert that important fields are not empty

sungchun12: By default, important fields will be validated to be non-empty. We can revisit if we see breakage in the wild because of this.

PR #694, opened 2023-09-06
add option `--init-oracle-client-lib-dir`

qsliu2017:
4b2ca27479d28e7409d0b01404154319072c165f switchs oracle driver from `cx_Oracle` to `oracledb`, causing older oracle databases (11g particularly) [not supported](https://python-oracledb.readthedocs.io/en/latest/user_guide/installation.html#supported-oracle-database-versions)

this PR add `--init-oracle-client-lib-dir` option to enable oracledb’s [thick mode](https://python-oracledb.readthedocs.io/en/latest/user_guide/installation.html#supported-oracle-database-versions), and therefore support oralce 11g

### DIFF
--- a/data_diff/__main__.py
+++ b/data_diff/__main__.py
@@ -250,6 +250,12 @@ click.Context.formatter_class = MyHelpFormatter
     help="Which directory to look in for the dbt_project.yml file. Default is the current working directory and its parents.",
 )
 @click.option(
+    "--init-oracle-client-lib-dir",
+    default=None,
+    metavar="PATH",
+    help="Path to the Oracle client library. Only relevant for Oracle databases.",
+)
+@click.option(
     "--select",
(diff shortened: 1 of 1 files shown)


data_diff/__main__.py
```diff
@@ -402,6 +409,16 @@ def _data_diff(
         )
         return
 
+    if init_oracle_client_lib_dir:
```
dlawin, line 412: I think that we should move this oracle related code to https://github.com/datafold/data-diff/blob/master/data_diff/sqeleton/databases/oracle.py

And it could be enabled by an environment variable instead. e.g. `LEGACY_ORACLE_LIB`

The main method here is already messy and don't want to make it worse

dlawin (changes requested):
see comment

PR #700, opened 2023-09-13
remove python 3.7 support

dlawin:


nolar (approved):

.github/workflows/ci_full.yml
```diff
@@ -21,7 +21,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         python-version:
-          - "3.10"
+          - "3.11"
```
nolar, line 24: Python 3.12 is planned to be release today, 2023-10-02 (probably US time). We can add the support for 3.12 too.

Monitor at:

* https://www.python.org/downloads/
* https://github.com/actions/python-versions/releases (the CI-enabled versions).

PR #706, opened 2023-09-22
Normalise all imports to fully qualified names

nolar:
In order to use the same convention across the code (now: mixed relative & absolute imports).

Besides, this helps to easily grep the imports, so as to use IDEs features to auto-import modules on use.

Besides, the import sorters typically convert this to fully qualified names anyway.

**THERE IS NO CHANGE IN FUNCTIONALITY.**


nolar:
Supersedes by #707 with the branch in the main repo, so that I could build a chain of PRs.

PR #707, opened 2023-09-22
Normalise all imports to fully qualified names

nolar:
In order to use the same convention across the code (now: mixed relative & absolute imports).

Besides, this helps to easily grep the imports, so as to use IDEs features to auto-import modules on use.

Besides, the import sorters typically convert this to fully qualified names anyway.

**THERE IS NO CHANGE IN FUNCTIONALITY.**

### DIFF
--- a/data_diff/cloud/datafold_api.py
+++ b/data_diff/cloud/datafold_api.py
@@ -9,8 +9,7 @@ import requests
 from typing_extensions import Self
 
 from data_diff.errors import DataDiffCloudDiffFailed, DataDiffCloudDiffTimedOut, DataDiffDatasourceIdNotFoundError
-
-from ..utils import getLogger
+from data_diff.utils import getLogger
 
 logger = getLogger(__name__)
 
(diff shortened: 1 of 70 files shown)


dlawin:
@nolar, looks good. I think we need to modify how pytest is being called in CI due to failures like `ModuleNotFoundError: No module named 'data_diff.cloud.utils'`

data_diff/cloud/datafold_api.py
```diff
@@ -10,7 +10,7 @@
 
 from data_diff.errors import DataDiffCloudDiffFailed, DataDiffCloudDiffTimedOut, DataDiffDatasourceIdNotFoundError
 
-from ..utils import getLogger
+from data_diff.cloud.utils import getLogger
```
dlawin, line 13: actually this should be data_diff.utils

dlawin:
@nolar looks like some of the relative import updates are incorrect. See my comment above

data_diff/cloud/datafold_api.py
nolar: solved

Issue #603, opened 2023-06-07
TypeError when running for MySQL


dschultz5197:
**Describe the bug**
Attempting to run data-diff for a MySQL table is resulting in `TypeError: [UnknownColType] Attribute 'text' expected value of type str. Instead got b'int'`

Below is output running the CLI.

```
~/code/test $ data-diff mysql://user:password@mysql_host:3306/database table_a table_b --key-columns=col_a --columns=col_b --debug
[14:03:57] INFO - Parsing DSN mysql://user:password@mysql_host:3306/database with ConnectionURI
[14:03:57] DEBUG - Parsed scheme mysql
[14:03:57] DEBUG - Parsed authority with 65 characters
[14:03:57] DEBUG - Parsed username user
[14:03:57] DEBUG - Parsed password with 6 characters
[14:03:57] DEBUG - Parsed host mysql_host with port 3306
[14:03:57] DEBUG - Parsed path /database
[14:03:57] DEBUG - Parsed query with 0 characters
[14:03:57] DEBUG - Parsed query_params with  keys
[14:03:57] INFO - [MySQL] Starting a threadpool, size=1.
[14:03:57] DEBUG - Running SQL (MySQL): SET @@session.time_zone='+00:00'
[14:03:57] DEBUG - # _do_auth(): user: user
[14:03:57] DEBUG - # _do_auth(): self._auth_plugin:
[14:03:57] DEBUG - # _do_auth(): user: user
[14:03:57] DEBUG - # _do_auth(): password: password
[14:03:57] DEBUG - new_auth_plugin: mysql_native_password
[14:03:57] DEBUG - Running SQL (MySQL): SELECT column_name, data_type, datetime_precision, numeric_precision, numeric_scale FROM information_schema.columns WHERE table_name = 'table_a' AND table_schema = 'database'
[14:03:57] DEBUG - Running SQL (MySQL): SELECT column_name, data_type, datetime_precision, numeric_precision, numeric_scale FROM information_schema.columns WHERE table_name = 'table_b' AND table_schema = 'database'
[...]

PR #711, opened 2023-09-22
add additional debug logs for mysql charset

dlawin:
I'm thinking issue #603 stems from ([source](https://dev.mysql.com/doc/refman/8.0/en/charset-connection.html#charset-connection-error-handling)):

> If you specify a character set that the client recognizes but the server does not, the server falls back to its default character set and collation.

I'd like to verify that by adding this logging.


dlawin:
actually I don't think this is the issue.

PR #720, opened 2023-09-29
Convert to `attrs`, remove `runtype`

nolar:
**Short-term goal:**  resolve an issue where `runtype` cannot check `isinstance()` check, where the type is a string `"Database"` coming from `Compilar.database: "Database"`. 

Technically, it is a `typing.ForwardRef[]` and must be resolved to actual classes on usage. `attrs` and `pydantic` do that.

There is no way to convert it to a regular class, because there is still a circular class dependency `Database` <> `Compiler`.

---

**Long-term goal:** Since we are moving in the direction of strict type checking, let's make the first step: replace `runtype` with lots of implicit logic and not compatible with MyPy, with `attrs`, which require explicitness and work fine with MyPy & IDEs.

`attrs` is much more beneficial:

* `attrs` is supported by type checkers, such as MyPy & IDEs
* `attrs` is widely used and industry-proven
* `attrs` is explicit in its declarations, there is no magic
* `attrs` has slots

But mainly for the first item — type checking by type checkers.

And since we switch to `attrs` for already dataclass-like types, also expand it to all other (unannotated) classes to ensure proper attribute management in multiple inheritance.

In particular, in the last commit, there are a few cases where the code was ambiguous and had to be resolved by minor remakes.

---

**It is better to review commit-by-commit.**

### DIFF
--- a/data_diff/abcs/compiler.py
+++ b/data_diff/abcs/compiler.py
@@ -1,9 +1,13 @@
 from abc import ABC
 
+import attrs
 
+
+@attrs.define
 class AbstractCompiler(ABC):
     pass
 
 
+@attrs.define(eq=False)
 class Compilable(ABC):
     pass
--- a/data_diff/abcs/database_types.py
+++ b/data_diff/abcs/database_types.py
@@ -15,7 +15,9 @@ DbTime = datetime
 
 @dataclass
 class ColType:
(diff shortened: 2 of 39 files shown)


nolar:
Accidentally closed by deleting the branch. Supersedes by #723.

PR #739, opened 2023-10-12
add query retries for postgres

dlawin:
Adds a new class ThreadedDatabaseWithRetry that can retry queries when worker threads are interrupted by temporary networking issues. Switches the Postgres connector to use it

### DIFF
--- a/data_diff/databases/postgresql.py
+++ b/data_diff/databases/postgresql.py
@@ -1,6 +1,7 @@
 from typing import Any, ClassVar, Dict, List, Type
 
 import attrs
+from psycopg2 import OperationalError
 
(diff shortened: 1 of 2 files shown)


data_diff/databases/postgresql.py
```diff
@@ -127,7 +128,7 @@ def type_repr(self, t) -> str:
 
 
 @attrs.define(frozen=False, init=False, kw_only=True)
-class PostgreSQL(ThreadedDatabase):
+class PostgreSQL(ThreadedDatabaseWithRetry):
```
dlawin, line 131: Redshift inherits PostgresSQL, self reminder to check/test this

PR #746, opened 2023-10-17
add checksum offset to avoid bigint overflow

vvkh:
We utilize a portion of the MD5 checksum, specifically 6 bytes per row, to allow for the accumulation of more rows before encountering overflow. However, in most databases, hex numbers are padded with zeros if they are not 8 bytes long. For example, `0xFF` becomes `0x00000000000000FF`.

Since the most significant bit is always 0, the number is consistently positive. Consequently, there is a limit to the number of values we can sum before encountering overflow in the bigint type.

To address this limitation, we introduce a negative offset equal to half of the maximum checksum number. This adjustment expands the range of values to "-max number/2" to "+max number/2". By incorporating negative numbers, we can offset some of the positive ones, enabling the summation of more numbers without encountering bigint overflow.

**Smoke tests for some databases after update**
Compute checksum from a string literal "hello"

| db         | expr                                                                                                                              | checksum       |
|------------|-----------------------------------------------------------------------------------------------------------------------------------|----------------|
| clickhouse | `reinterpretAsUInt128(reverse(unhex(lowerUTF8(substr(hex(MD5('hello')), 21))))) - 140737488355327`                                | 32508877456787 |
| bigquery   | `cast(cast(('0x' \|\| substr(TO_HEX(md5("hello")), 21)) as int64) as numeric) - 140737488355327`                                  | 32508877456787 |
| snowflake  | `BITAND(md5_number_lower64('hello'), 281474976710655) - 140737488355327`                                                          | 32508877456787 |
| postgres   | `('x' \|\| substring(md5('hello'), 21))::bit(48)::bigint - 140737488355327`                                                       | 32508877456787 |
| mssql      | `convert(bigint, convert(varbinary, '0x' + RIGHT(CONVERT(NVARCHAR(32), HashBytes('MD5', 'hello'), 2), 12), 1)) - 140737488355327` | 32508877456787 |
| mysql      | `conv(substring(md5('hello'), 21), 16, 10) - 140737488355327`                                                                     | 32508877456787 |
| oracle     | `to_number(substr(standard_hash('hello', 'MD5'), 21), 'xxxxxxxxxxxxxxx') - 140737488355327`                                       | 32508877456787 |

### DIFF
--- a/data_diff/databases/bigquery.py
+++ b/data_diff/databases/bigquery.py
@@ -37,7 +37,7 @@ from data_diff.databases.base import (
     apply_query,
     QueryResult,
     CHECKSUM_OFFSET,
-    CHECKSUM_HEXDIGITS
+    CHECKSUM_HEXDIGITS, MD5_HEXDIGITS
 )
 from data_diff.databases.base import TIMESTAMP_PRECISION_POS, ThreadLocalInterpreter, Mixin_RandomSample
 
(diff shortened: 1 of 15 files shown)


data_diff/databases/bigquery.py
```diff
@@ -62,7 +65,7 @@ def import_bigquery_service_account_impersonation():
 @attrs.define(frozen=False)
 class Mixin_MD5(AbstractMixin_MD5):
     def md5_as_int(self, s: str) -> str:
-        return f"cast(cast( ('0x' || substr(TO_HEX(md5({s})), 18)) as int64) as numeric)"
```
vvkh, line 65: The number of digits used from the checksum was hardcoded and not the same as in other databases 🫠

nolar (approved):

PR #756, opened 2023-10-28
feat: support TOP operator

pik94:
Some databases does not support `LIMIT` operator but has `TOP`. This PR adds supporting of `TOP`. A flag `USE_TOP_INSTEAD_LIMIT` controls which op should be used for a specific database.


dlawin:
offset_limit() must not be enough? ex: https://github.com/datafold/data-diff/blob/master/data_diff/databases/mssql.py#L113C9-L113C21

dlawin (approved):

data_diff/databases/base.py
```diff
@@ -471,7 +472,10 @@ def render_select(self, parent_c: Compiler, elem: Select) -> str:
         columns = ", ".join(map(compile_fn, elem.columns)) if elem.columns else "*"
         distinct = "DISTINCT " if elem.distinct else ""
         optimizer_hints = self.optimizer_hints(elem.optimizer_hints) if elem.optimizer_hints else ""
-        select = f"SELECT {optimizer_hints}{distinct}{columns}"
+        if elem.limit_expr is not None and self.USE_TOP_INSTEAD_LIMIT:
+            select = f"SELECT TOP {elem.limit_expr} {optimizer_hints}{distinct}{columns}"
+        else:
+            select = f"SELECT {optimizer_hints}{distinct}{columns}"
```
nolar, line 478: Having this part of SQL rendering done via `if` instead of the usual AST-based & `def render_…()` approach looks unreliable.

Can I propose a little refactoring/redesign to make (maybe as a separate PR)?

Having a fully rendered SELECT SQL in the end of this routine, send it to some new function called e.g. `limit_select(sql: str)` (a better name should be picked), which will **wrap** the provided SQL into the following one by default:

```sql
SELECT * FROM ({sql}) LIMIT {n}
```

And overrides it in the SELECT-TOP-enabled databases to be:

```sql
SELECT TOP {n} FROM ({sql})
```

We also had a few cases with unusual LIMIT syntax, which is nicely solved by this approach:

MS SQL:

```sql
SELECT * FROM ({sql}) OFFSET 0 ROWS FETCH NEXT {n} ROWS ONLY
```

Oracle:

```sql
SELECT * FROM ({sql}) FETCH NEXT {n} ROWS ONLY
```

I.e., limit the selection from a subquery instead of modifying the query itself. All databases, to my best knowledge, can optimize such cases and do not fetch the full table before limiting it.

Then remove `def offset_limit(…)` as unused & unneeded.

nolar (changes requested):

data_diff/databases/base.py
pik94: @nolar could you clarify why unreliable?

I thought about adding something like this but i might affect performance negatively because of subquering for limit. Right now there are no many places where `limit` is actively used but to avoid any problems in the future i decided to implement it like this.

If we still want not to use `if` we can try using @dlawin approach.

nolar: `offset_limit()` would be a problem because the `TOP {n}` is injected somewhere at the top of SQL, while that offset-limit thing implies one specific place in the end.

Regarding performance issues: are there any known cases when the database server cannot optimize a subquery as simple as that? Unwrapping subqueries into the optimizer was the capability of databases since maybe the early 2000s, if not the 1990s.

Regarding unreliable — it **seems** so, not necessary it **is** so. Maybe I'm overthinking. I'll try to explain my view:

Referring to currently trendy "best practices", with some interpretation, this breaks maybe half of the SOLID principles. But SOLID itself is not the best practice as it leads to overengineering (in Python, at least).

What causes my concerns on a bigger scale is a more specific thing: we chose to define dialects via polymorphism, i.e., with method overrides. Adding the flag-based approach breaks this principle and furthers us into an all-in-one "god function" (which the SELECT SQL rendering already is). Now, we have 2 methods of overriding the behavior: the polymorphic and the flag-based one. First, it is not nice. Second, based on experience, mixed approaches in code design always lead to problems eventually in the future (and we will later learn yet another new lesson why best practices exist, but we do not know which one at the moment).

On a smaller/more specific scale, I am worried because it brings specifics of one and only one rare dialect into the base class of all dialects, i.e., it puts an ad-hoc solution into the generalized base class. That case opens up a road to putting all other dialects into the base classes via `if`s, ruining the whole idea of using OOP & polymorphism for this purpose (a lot of such cases have been removed recently to simplify the codebase).

But overall, if the suggested refactoring is risky & problematic, this ad-hoc solution would work too — until we see an emerging pattern with more dialects.

Please also make sure in tests that this TOP-limit for this database is not lost on further accidental changes.

pik94: @nolar i see, thank you for such a detailed clarification. Your args sounds rational and logical. I have rewritten the previous implementation adding a new method `limit_select` and removing `offset_limit`.

dlawin (approved):

nolar (approved):

PR #757, opened 2023-10-28
Prevent type overflow

pik94:
Some databases, for example, Teradata, have a short limit for string and requires to set a number of symbols in `CHAR/VARCHAR`, i.e. it is required to write it like this
```
CREATE TABLE MyTable (
    col_name VARCHAR(n)
)
```
where `n` must be set and has some max allowed value.

To calculate md5 hash and convert it to an integer value, values of columns for one row are cast to strings, and concatenated afterwards. Let me clarify on an example:
```
CREATE TABLE MyTable (
    id INTEGER,
    data VARCHAR(n)
)
```
To concatenate we use a construction like this:
```
CAST(id AS VARCHAR(n1)) || '|'  || CAST(data AS VARCHAR(n2)) 
``` 
The question is what should `n1` and `n2` be? I am sure that the maximum allowed value for a specific type, for example, `N`. It is needed to keep all customer information without losses. However, such a concatenation will lead to a type overflow, because we are trying to have `VARCHAR(N + N)` which is not allowed.

To avoid such an overflow, we should shorten string values but not to loss information. I see one possible solution: taking hash for each item of a concat op, i.e.
```
md5( CAST(id AS VARCHAR(n1)) ) || '|'  || md5( CAST(data AS VARCHAR(n2)) )
``` 

Benefits:
- It is possible to avoid data type overflow for most of cases

Drawbacks:
- Performance might be decreased because we need to calculate hashes quite intensively. However, the current implementation enables it if necessarily i.e. only if at least one of databases participating in cross-diffing has a problem with a fixed varchar. This behavior is controlled by a `PREVENT_OVERFLOW_WHEN_CONCAT` flag.
- Probability of collision increases but i do not think dramatically.
- We still may have data overflow if where are many columns and a max `n` of `VARCHAR(n)` is low. Do not think it is a big problem because typically a large `N` might be 32000 or more symbols so customer columns should have more than 1000 column in diffing.

### DIFF
--- a/data_diff/databases/base.py
+++ b/data_diff/databases/base.py
@@ -403,7 +403,8 @@ class BaseDialect(abc.ABC):
     def render_concat(self, c: Compiler, elem: Concat) -> str:
         if self._prevent_overflow_when_concat:
             items = [
-                f"{self.compile(c, Code(self.md5_as_hex(self.to_string(self.compile(c, expr)))))}" for expr in elem.exprs
+                f"{self.compile(c, Code(self.md5_as_hex(self.to_string(self.compile(c, expr)))))}"
+                for expr in elem.exprs
             ]
 
         # We coalesce because on some DBs (e.g. MySQL) concat('a', NULL) is NULL
--- a/data_diff/databases/base.py
+++ b/data_diff/databases/base.py
@@ -203,6 +203,15 @@ class BaseDialect(abc.ABC):
 
     PLACEHOLDER_TABLE = None  # Used for Oracle
 
+    # Some database do not support long string so concatenation might lead to type overflow
+    PREVENT_OVERFLOW_WHEN_CONCAT: bool = False
+
+    _prevent_overflow_when_concat: bool = False
+
+    def enable_preventing_type_overflow(self) -> None:
+        logger.info("Preventing type overflow when concatenation is enabled")
+        self._prevent_overflow_when_concat = True
+
     def parse_table_name(self, name: str) -> DbPath:
         "Parse the given table name into a DbPath"
         return parse_table_name(name)
@@ -392,10 +401,18 @@ class BaseDialect(abc.ABC):
         return f"sum({md5})"
 
     def render_concat(self, c: Compiler, elem: Concat) -> str:
+        if self._prevent_overflow_when_concat:
+            items = [
+                f"{self.compile(c, Code(self.to_md5(self.to_string(self.compile(c, expr)))))}" for expr in elem.exprs
+            ]
+
         # We coalesce because on some DBs (e.g. MySQL) concat('a', NULL) is NULL
-        items = [
-            f"coalesce({self.compile(c, Code(self.to_string(self.compile(c, expr))))}, '<null>')" for expr in elem.exprs
-        ]
+        else:
+            items = [
+                f"coalesce({self.compile(c, Code(self.to_string(self.compile(c, expr))))}, '<null>')"
+                for expr in elem.exprs
+            ]
+
         assert items
         if len(items) == 1:
             return items[0]
@@ -768,6 +785,10 @@ class BaseDialect(abc.ABC):
     @abstractmethod
     def md5_as_int(self, s: str) -> str:
         "Provide SQL for computing md5 and returning an int"
+
+    @abstractmethod
+    def to_md5(self, s: str) -> str:
+        """Method to calculate MD5"""
 
     @abstractmethod
     def normalize_timestamp(self, value: str, coltype: TemporalType) -> str:
--- a/data_diff/databases/base.py
+++ b/data_diff/databases/base.py
@@ -906,6 +906,8 @@ class Database(abc.ABC):
     Instanciated using :meth:`~data_diff.connect`
     """
 
+    DIALECT_CLASS: ClassVar[Type[BaseDialect]] = BaseDialect
+
     SUPPORTS_ALPHANUMS: ClassVar[bool] = True
     SUPPORTS_UNIQUE_CONSTAINT: ClassVar[bool] = False
     CONNECT_URI_KWPARAMS: ClassVar[List[str]] = []
@@ -913,6 +915,7 @@ class Database(abc.ABC):
     default_schema: Optional[str] = None
     _interactive: bool = False
     is_closed: bool = False
+    _dialect: BaseDialect = None
 
     @property
     def name(self):
@@ -1141,9 +1144,12 @@ class Database(abc.ABC):
         return super().close()
 
     @property
-    @abstractmethod
     def dialect(self) -> BaseDialect:
         "The dialect of the database. Used internally by Database, and also available publicly."
+
+        if not self._dialect:
+            self._dialect = self.DIALECT_CLASS()
+        return self._dialect
 
     @property
     @abstractmethod
--- a/data_diff/databases/base.py
+++ b/data_diff/databases/base.py
@@ -403,7 +403,7 @@ class BaseDialect(abc.ABC):
     def render_concat(self, c: Compiler, elem: Concat) -> str:
         if self._prevent_overflow_when_concat:
             items = [
-                f"{self.compile(c, Code(self.to_md5(self.to_string(self.compile(c, expr)))))}" for expr in elem.exprs
+                f"{self.compile(c, Code(self.md5_as_hex(self.to_string(self.compile(c, expr)))))}" for expr in elem.exprs
             ]
 
         # We coalesce because on some DBs (e.g. MySQL) concat('a', NULL) is NULL
@@ -787,7 +787,7 @@ class BaseDialect(abc.ABC):
         "Provide SQL for computing md5 and returning an int"
 
     @abstractmethod
-    def to_md5(self, s: str) -> str:
+    def md5_as_hex(self, s: str) -> str:
         """Method to calculate MD5"""
 
     @abstractmethod
--- a/data_diff/databases/base.py
+++ b/data_diff/databases/base.py
@@ -199,12 +199,12 @@ def apply_query(callback: Callable[[str], Any], sql_code: Union[str, ThreadLocal
 class BaseDialect(abc.ABC):
     SUPPORTS_PRIMARY_KEY: ClassVar[bool] = False
     SUPPORTS_INDEXES: ClassVar[bool] = False
+    PREVENT_OVERFLOW_WHEN_CONCAT: ClassVar[bool] = False
     TYPE_CLASSES: ClassVar[Dict[str, Type[ColType]]] = {}
 
     PLACEHOLDER_TABLE = None  # Used for Oracle
 
     # Some database do not support long string so concatenation might lead to type overflow
-    PREVENT_OVERFLOW_WHEN_CONCAT: bool = False
 
     _prevent_overflow_when_concat: bool = False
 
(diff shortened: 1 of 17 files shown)


data_diff/databases/base.py
```diff
@@ -202,6 +202,16 @@ class BaseDialect(abc.ABC):
     TYPE_CLASSES: ClassVar[Dict[str, Type[ColType]]] = {}
 
     PLACEHOLDER_TABLE = None  # Used for Oracle
+    USE_TOP_INSTEAD_LIMIT: bool = False  # True for MsSQL or Teradata
+
+    # Some database do not support long string so concatenation might lead to type overflow
+    PREVENT_OVERFLOW_WHEN_CONCAT: bool = False
+
+    _prevent_overflow_when_concat: bool = False
```
dlawin, line 210: did you mean to have two of these variables?

data_diff/databases/base.py
```diff
@@ -392,10 +402,18 @@ def render_checksum(self, c: Compiler, elem: Checksum) -> str:
         return f"sum({md5})"
 
     def render_concat(self, c: Compiler, elem: Concat) -> str:
+        if self._prevent_overflow_when_concat:
+            items = [
+                f"{self.compile(c, Code(self.to_md5(self.to_string(self.compile(c, expr)))))}" for expr in elem.exprs
```
dlawin, line 407: render_concat/Concat in general isn't specific to hashing, so not sure this is the best place to add `self.to_md5`

dlawin: To me, this method is pretty generic, so I just have some concerns about introducing logic specific to hashing here

data_diff/databases/base.py
```diff
@@ -471,7 +489,10 @@ def render_select(self, parent_c: Compiler, elem: Select) -> str:
         columns = ", ".join(map(compile_fn, elem.columns)) if elem.columns else "*"
         distinct = "DISTINCT " if elem.distinct else ""
         optimizer_hints = self.optimizer_hints(elem.optimizer_hints) if elem.optimizer_hints else ""
-        select = f"SELECT {optimizer_hints}{distinct}{columns}"
+        if elem.limit_expr is not None and self.USE_TOP_INSTEAD_LIMIT:
+            select = f"SELECT TOP {elem.limit_expr} {optimizer_hints}{distinct}{columns}"
+        else:
+            select = f"SELECT {optimizer_hints}{distinct}{columns}"
```
nolar, line 495: Having this part of SQL rendering done via `if` instead of the usual AST-based & `def render_…()` approach looks unreliable.

Can I propose a little refactoring/redesign to make (maybe as a separate PR)?

Having a fully rendering SELECT SQL in the end of this routine, send it to some new function called e.g. `limit_select(sql: str)` (a better name should be picked), which will **wrap** the provided SQL into the following one by default:

```sql
SELECT * FROM ({sql}) LIMIT {n}
```

And overrides it in the SELECT-TOP-enabled databases to be:

```sql
SELECT TOP {n} FROM ({sql})
```

We also had a few cases with unusual LIMIT syntax, which is nicely solved by this approach:

MS SQL:

```sql
SELECT * FROM ({sql}) OFFSET 0 ROWS FETCH NEXT {n} ROWS ONLY
```

Oracle:

```sql
SELECT * FROM ({sql}) FETCH NEXT {n} ROWS ONLY
```

I.e., limit the selection from a subquery instead of modifying the query itself. All databases, to my best knowledge, can optimize such cases and do not fetch the full table before limiting it.

Then remove `def offset_limit(…)` as unused & unneeded.

data_diff/databases/oracle.py
```diff
@@ -133,6 +133,9 @@ def md5_as_int(self, s: str) -> str:
         # TODO: Find a way to use UTL_RAW.CAST_TO_BINARY_INTEGER ?
         return f"to_number(substr(standard_hash({s}, 'MD5'), {1+MD5_HEXDIGITS-CHECKSUM_HEXDIGITS}), 'xxxxxxxxxxxxxxx') - {CHECKSUM_OFFSET}"
 
+    def to_md5(self, s: str) -> str:
+        return f"standard_hash({s}, 'MD5'"
```
nolar, line 137: Is the closing parenthesis lost here?

data_diff/databases/base.py
```diff
@@ -766,6 +787,10 @@ def set_timezone_to_utc(self) -> str:
     def md5_as_int(self, s: str) -> str:
         "Provide SQL for computing md5 and returning an int"
 
+    @abstractmethod
+    def to_md5(self, s: str) -> str:
```
nolar, line 791: May I suggest the name `md5_as_hex()` — to be consistent with `md5_as_int()`?

data_diff/diff_tables.py
```diff
@@ -208,6 +208,10 @@ def _diff_tables_wrapper(self, table1: TableSegment, table2: TableSegment, info_
             event_json = create_start_event_json(options)
             run_as_daemon(send_event_json, event_json)
 
+        if table1.database.dialect.PREVENT_OVERFLOW_WHEN_CONCAT or table2.database.dialect.PREVENT_OVERFLOW_WHEN_CONCAT:
+            table1.database.dialect.enable_preventing_type_overflow()
+            table2.database.dialect.enable_preventing_type_overflow()
```
nolar, line 213: ❗️This would be problematic. 

Sadly, at this moment, dialect is a global class-level object. And we reuse it for multiple runs across multiple databases. If it is mutated here for a pair of databases X-Y, it will also affect other pairs X-Z & Y-Z by breaking them — because the upper case constants are not modified, only the lower-case fields are, so the hashing logic will be different.

Restoring the fields to their original values will also not be reliable, as some xdb diffs can be prematurely terminated by exceptions and the workers reused after that for other xdb diffs — catching the issues with the modified dialect.

A good solution would be to make `database.dialect` a property that is created per database object and discarded afterwards. This is on the roadmap anyway for different reasons, but not a priority. 

Though it seems as easy as doing:

```python
@attrs.define
class WhateverDatabase:
    dialect: BaseDialect = attrs.field(factory=WhateverDialect)
```

Instead of the current

```python
@attrs.define
class WhateverDatabase:
    dialect = WhateverDialect()
```

Can you please try it?

A less good way would be to store the state somewhere else, e.g. in databases or in ColTypes (created for each run and discarded afterwards).

nolar (changes requested):

data_diff/databases/oracle.py
pik94: Yep, thank you!

data_diff/diff_tables.py
pik94: It is a good catch, thank you very much. I just thought that `attrs.define` makes as the same thing as dataclass but we still have a classvar here. I will try to fix it using your suggestion.

data_diff/databases/base.py
pik94: Yep, i though to introduce this param as internal but it is still classvar so it is not a good option. I will rewrite it.

data_diff/diff_tables.py
pik94: @nolar your solution does not work.

nolar: attrs works similar to dataclasses: if you want a classvar, annotate its type as `varname: ClassVar[BaseDialect] = ...` — in that case, attrs will ignore it entirely. The same as when there is no type annotation for a field, but that is not nice.

>  Cons here are it is needed to create it each time when dialect is accessed and implement the property for each connector.

In that case, you will lose its state, including the overflow prevention flag — the `db.dialect` is accessed many times at different occasions.

> your solution does not work.

What is the exact problem? Error?

pik94: > In that case, you will lose its state, including the overflow prevention flag — the db.dialect is accessed many times at different occasions.

Yes, seems true

> What is the exact problem? Error?

For example, if i run it postgres vs postgres, the error is
`AttributeError: 'PostgreSQL' object has no attribute 'dialect'`

pik94: I managed to emulate it in a script:
```
import attrs
import abc


@attrs.define(frozen=False)
class BaseParam:
    _flag_enabled: bool = False

    def enable(self):
        self._flag_enabled = True

    def print(self):
        print('base param')


@attrs.define(frozen=False)
class Param(BaseParam):
    def print(self):
        print('param')


@attrs.define(frozen=False, kw_only=True)
class Base(abc.ABC):
    var: int = 1

    @property
    @abc.abstractmethod
    def param(self) -> BaseParam:
        pass


@attrs.define(frozen=False, kw_only=True)
class MyClass(Base):
    param: BaseParam = attrs.field(factory=Param)
    _args: dict[str, any]

    def __init__(self, *, var, **kw):
        super().__init__(var=var)
        self._args = kw


if __name__ == '__main__':
    kw = {'a': 1, 'b': 2}
    c1 = MyClass(var=1, **kw)
    c2 = MyClass(var=1, **kw)
    c1.param.print()
```

data_diff/databases/base.py
pik94: I tried to find another way but looks like the current implementation is the right place because this part is common for all databases, and only here it is possible to prevent overflow.

data_diff/diff_tables.py
pik94: @nolar 
i managed to rewrite it to use the local instance of the dialect class instead of class var and pushed it. The short idea in the script below:
```
import attrs
import abc


@attrs.define(frozen=False)
class BaseParam:
    _flag_enabled: bool = False

    def enable(self):
        self._flag_enabled = True

    def print(self):
        print('base param')


@attrs.define(frozen=False)
class Param(BaseParam):
    def print(self):
        print('param')


@attrs.define(frozen=False, kw_only=True)
class Base(abc.ABC):
    Class = BaseParam
    var: int = 1
    _param: BaseParam = None

    @property
    def param(self) -> BaseParam:
        if not self._param:
            self._param = self.Class()
        return self._param


@attrs.define(frozen=False, kw_only=True)
class MyClass(Base):
    Class = Param
    _args: dict[str, any]

    def __init__(self, *, var, **kw):
        super().__init__(var=var)
        self._args = kw


if __name__ == '__main__':
    kw = {'a': 1, 'b': 2}
    c1 = MyClass(var=1, **kw)
    c2 = MyClass(var=1, **kw)
    c1.param.print()
    print(c1._param)
```

data_diff/databases/base.py
```diff
@@ -471,7 +490,10 @@ def render_select(self, parent_c: Compiler, elem: Select) -> str:
         columns = ", ".join(map(compile_fn, elem.columns)) if elem.columns else "*"
         distinct = "DISTINCT " if elem.distinct else ""
         optimizer_hints = self.optimizer_hints(elem.optimizer_hints) if elem.optimizer_hints else ""
-        select = f"SELECT {optimizer_hints}{distinct}{columns}"
+        if elem.limit_expr is not None and self.USE_TOP_INSTEAD_LIMIT:
+            select = f"SELECT TOP {elem.limit_expr} {optimizer_hints}{distinct}{columns}"
+        else:
+            select = f"SELECT {optimizer_hints}{distinct}{columns}"
```
dlawin, line 496: are these changes still intended to exist in this PR? The approach has changed in #756

pik94: Updated.

dlawin (approved):

PR #763, opened 2023-11-02
DX-762 - initial attempt at a fix

mrn3:


ramdharam:
@mrn3 - you could add 
`alphanums = string.digits + string.ascii_lowercase + string.ascii_uppercase` in your data_diff/utils.py 

That will give you a bigger range when you convert strings to numbers and break the table into equal sized chunks.

Issue #750, opened 2023-10-24
data-diff fails when dbt anonymous usage is disabled


eschrock:
**Describe the bug**

Running `data-diff` fails:

```
Running with data-diff=0.9.7
04:09:46 ERROR    1 validation error for ManifestJsonConfig                                                                                                                     __main__.py:332
                  metadata -> user_id                                                                                                                                                          
                    none is not an allowed value (type=type_error.none.not_allowed)                                                                                                            
Traceback (most recent call last):
  File "/usr/local/bin/data-diff", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.10/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/data_diff/__main__.py", line 315, in main
    dbt_diff(
  File "/usr/local/lib/python3.10/site-packages/data_diff/dbt.py", line 83, in dbt_diff
    dbt_parser = DbtParser(profiles_dir_override, project_dir_override, state)
  File "/usr/local/lib/python3.10/site-packages/data_diff/dbt_parser.py", line 128, in __init__
    self.dev_manifest_obj = self.get_manifest_obj(self.project_dir / MANIFEST_PATH)
  File "/usr/local/lib/python3.10/site-packages/data_diff/dbt_parser.py", line 282, in get_manifest_obj
    manifest_obj = ManifestJsonConfig.parse_obj(manifest_dict)
  File "pydantic/main.py", line 526, in pydantic.main.BaseModel.parse_obj
  File "pydantic/main.py", line 341, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 1 validation error for ManifestJsonConfig
metadata -> user_id
  none is not an allowed value (type=type_error.none.not_allowed)
```

Sure enough, looking at the pydantic model in data-diff the field is not declared optional:

```
class ManifestJsonConfig(BaseModel):
    class Metadata(BaseModel):
[...]

PR #769, opened 2023-11-04
tracking should be optional in dbt configuration

eschrock:
This addresses the issue #750, which is caused from the dbt config validator requiring `user_id` (as well as `project_id`). These can be null when anonymous tracking is disabled. The rest of the code treats this as optional, so the fix is just addressing the type in the pydantic model. I added a new flavor of a v11 manifest for testing, and verified that this test failed with the previous code but succeeds with the new code. This is an identical copy to v11.json but has `project_name`, `project_id`, and `user_id` removed, and `send_anonymous_usage_stats` set to null.  I also verified that I could now run data-diff on the original project where it was failing.

Note that I couldn't get a completely clean test run due to some issues parsing dbt profiles that I posted about in the slack channel. However, those errors were present on the trunk code prior to the change.


tests/test_dbt_config_validators.py
```diff
@@ -26,7 +26,7 @@ def test_run_results(self):
 class TestManifestJsonConfig(unittest.TestCase):
     def test_manifest(self):
         # https://docs.getdbt.com/reference/artifacts/manifest-json
-        versions = ["v4", "v5", "v6", "v7", "v8", "v9", "v10", "v11"]
+        versions = ["v4", "v5", "v6", "v7", "v8", "v9", "v10", "v11", "v11_no_tracking" ]
```
sungchun12, line 29:
```suggestion
        versions = ["v4", "v5", "v6", "v7", "v8", "v9", "v10", "v11", "v11_no_tracking"]
```

sungchun12:
Thanks for this and your first time contributing to `data-diff`! Straightforward change. It means so much!

sungchun12 (approved):

PR #771, opened 2023-11-07
Add pre-commit automation for faster, safer development

sungchun12:
The goal of this PR is to remove the friction contributing to `data-diff` both locally and in PRs.

**Problem**: black formatting in the PR adds a bunch of annotations and run time that adds more friction than needed in the pull request workflow. It forces the contributor to click a lot of buttons and run black locally for those, "Oh shoot. I forgot to format!" moments. This adds a couple minutes at a time of frustration which over the course of a month, is a lot of frustration. 

**Solution**: 
- Added pre-commit configs to automatically enforce ruff formatting in local development after a contributor runs: `poetry install` and `pre-commit install`. Ensured line length is constrained to a max line length of 120 given the previous black settings. 
image
- Updated the formatter github action to use ruff (99.9% black compatible) which runs faster AND auto commits changes if they exist.

This wombo combo gives precious time and energy back to the contributor and declutters PRs of those annotations that should just be auto-committed.


dlawin (commented):
Approved but want to give @nolar a chance to look at it as well

dlawin (approved):

Issue #426, opened 2023-03-01
Run diffs in paralllel


dlawin:
**Is your feature request related to a problem? Please describe.**
Currently --dbt diffs run sequentially when there are multiple models in the last dbt run. This can cause long running diffs to slow down the overall total diff run time.

**Describe the solution you'd like**
Run the diffs for each model in parallel. Output as they finish.

PR #770, opened 2023-11-07
Add Diff Time after each Data Diff

sungchun12:
Resolves #672 

Let's make it easy for data differs to predict their diffing times so they when it's worth waiting for the command to run with patience or walking away from the computer to do better things with their time.

Working Draft
image

PR #776, opened 2023-11-13
Make local dbt data diffs concurrent

sungchun12:
In the midst of working on #770, I realized all local dbt data diffs run sequentially. This makes data-diff run slower than necessary to reach the same results. Race conditions shouldn't be an issue because each query `data-diff` runs are independent by design. This is also validated by the fact that cloud data diffs have concurrent diffs already built for awhile and have not encountered bugs. 

- Updated the concurrent thread mechanism using python's built-in `ThreadPoolExecutor` for cloud diffs and local diffs
- Added concurrent progress trackers similar to how cloud diffs track concurrent diffs
- Updated functions and debug logs with specificity to troubleshoot interleaved logging
- Added detailed tracebacks for errors
- No major tests added given most of the changes are debug log messages as these are auxiliary components rather than main business logic.
- Updated existing test logic to mock threads

**Before this PR:** Sequential performance is slow
image

**After this PR:** Consistently get 2-4x performance increase with concurrency enabled
image
image

[data-diff PR Demo V1 - Watch Video](https://www.loom.com/share/f25c442435034310838f1b4d5e39037a)
[Verify Concurrent Data Diffs work with VS Code Extension - Watch Video](https://www.loom.com/share/09ae032715e74b32904240dbae56825f)


data_diff/dbt.py
```diff
@@ -265,7 +266,11 @@ def _get_prod_path_from_manifest(model, prod_manifest) -> Union[Tuple[str, str,
     return prod_database, prod_schema, prod_alias
 
 
-def _local_diff(diff_vars: TDiffVars, json_output: bool = False) -> None:
+def _local_diff(
+    diff_vars: TDiffVars, json_output: bool = False, log_status_handler: Optional[LogStatusHandler] = None
+) -> None:
+    if log_status_handler:
+        log_status_handler.diff_started(diff_vars.dev_path[-1])
```
dlawin, line 273: In this method we're passing `diff_vars.threads` to connect_to_table, which is the `threads:` value from the dbt profile. This was \~okay\~ when running the diffs sequentially, but could greatly balloon the number of concurrent database threads when the outer diffs execute in parallel

Previously:
`n_concurrent_db_threads = diff_vars.threads`

PR as is:
`n_concurrent_db_threads = number_of_daemons * diff_vars.threads`

Running these in parallel, I think it makes more sense to use that value to handle the number of daemons running above:
```
diff_thread = run_as_daemon(_local_diff, diff_vars, json_output, log_status_handler)
diff_threads.append(diff_thread)
```

That would use the diff_vars.threads value more similarly to dbt. Having threads control the number of concurrent diffs.

sungchun12: So you want me to refactor `connect_to_table` to NOT use `diff_vars.threads` anymore and bring that up a level so that `threads==number_of_daemons`?

dlawin:
Resolves #426

data_diff/dbt.py
sungchun12: Removed `diff_vars.threads` param from `connect_to_table` and made it a param for max workers.
https://github.com/datafold/data-diff/pull/776/files#diff-8a5550cee0de299f3906d3b40b82bec419ce23c065b04923c3638906b21bda70R119

data_diff/dbt.py
```diff
@@ -165,10 +170,13 @@ def dbt_diff(
                         + "Skipped due to unknown primary key. Add uniqueness tests, meta, or tags.\n"
                     )
 
-        # wait for all threads
-        if diff_threads:
-            for thread in diff_threads:
-                thread.join()
+    for future in as_completed(futures):
```
sungchun12, line 173: Example Error log
image

sungchun12: I removed this as I realized it's too noisy to be helpful and that a user can just as easily add the `--debug` flag to figure out the problem.

image

image

data_diff/databases/base.py
```diff
@@ -931,7 +931,7 @@ def name(self):
     def compile(self, sql_ast):
         return self.dialect.compile(Compiler(self), sql_ast)
 
-    def query(self, sql_ast: Union[Expr, Generator], res_type: type = None):
+    def query(self, sql_ast: Union[Expr, Generator], res_type: type = None, log_message: str = None):
```
dlawin, line 934:
```suggestion
    def query(self, sql_ast: Union[Expr, Generator], res_type: type = None, log_message: Optional[str] = None):
```

dlawin (approved):

data_diff/joindiff_tables.py
```diff
@@ -334,23 +334,46 @@ def _create_outer_join(self, table1, table2):
         diff_rows = all_rows.where(or_(this[c] == 1 for c in is_diff_cols))
         return diff_rows, a_cols, b_cols, is_diff_cols, all_rows
 
-    def _count_diff_per_column(self, db, diff_rows, cols, is_diff_cols):
-        logger.debug("Counting differences per column")
-        is_diff_cols_counts = db.query(diff_rows.select(sum_(this[c]) for c in is_diff_cols), tuple)
+    def _count_diff_per_column(
```
sungchun12, line 337: added optional types for these params inspired by Dan's earlier commit suggestion

data_diff/joindiff_tables.py
```diff
@@ -334,23 +334,46 @@ def _create_outer_join(self, table1, table2):
...
+            tuple,
+            log_message=f"{table1.table_path} <> {table2.table_path}",
+        )
         diff_counts = {}
         for name, count in safezip(cols, is_diff_cols_counts):
             diff_counts[name] = diff_counts.get(name, 0) + (count or 0)
         self.stats["diff_counts"] = diff_counts
 
-    def _sample_and_count_exclusive(self, db, diff_rows, a_cols, b_cols):
+    def _sample_and_count_exclusive(
```
sungchun12, line 358: added optional types for these params inspired by Dan's earlier commit suggestion

data_diff/dbt.py
```diff
@@ -165,10 +169,12 @@ def dbt_diff(
...
+        model = futures[future]
```
sungchun12, line 173: Ensure model is mapped to the correct future result that caused the error

PR #785, opened 2023-11-28
closing connection once data diff is executed

sar009:
This PR revamps how we close database connections, adds new unit tests to the main entrypoint logic, and refactors tests for better readability.


sar009:
fixes issue #785

tests/test_postgresql.py
```diff
@@ -127,20 +128,22 @@ def setUp(self) -> None:
...
         password_quoted = quote(password)
-        db_config = {
-            "driver": "postgresql",
-            "host": "localhost",
-            "port": 5432,
-            "dbname": "postgres",
-            "user": "test",
-            "password": password_quoted,
-        }
+        db_config = deepcopy(self.connection._args)
```
sungchun12, line 138:
```suggestion
```
We don't need to complicate this config

tests/test_postgresql.py
```diff
@@ -127,20 +128,22 @@ def setUp(self) -> None:
...
+        db_config.update(
```
sungchun12, line 139: Make this config explicit

db_config = {
            "driver": "postgresql",
            "host": "localhost",
            "port": 5432,
            "dbname": "postgres",
            "user": username,
            "password": password_quoted,
        }

tests/test_postgresql.py
```diff
@@ -149,7 +152,10 @@ def test_special_char_password(self):
         )
 
         # verify connection method with URL string unquoted after it's verified
-        db_url = f"postgresql://{db_config['user']}:{db_config['password']}@{db_config['host']}:{db_config['port']}/{db_config['dbname']}"
+        db_url = (
+            f"postgresql://{db_config['user']}:{db_config['password']}@{db_config['host']}:"
+            f"{db_config.get('port', 5432)}/{db_config['dbname']}"
```
sungchun12, line 157: Revert this line to the original once you make the above db_config changes

tests/test_postgresql.py
```diff
@@ -1,11 +1,12 @@
 import unittest
-
+from copy import deepcopy
```
sungchun12, line 2:
```suggestion
```

sungchun12:
I'll review this later this week!

tests/test_postgresql.py
sar009: if explicit this test fails since it is not honouring configuration in `TEST_POSTGRESQL_CONN_STRING`.

sungchun12: ahh fair callout. I'll resolve this then.

data_diff/__main__.py
```diff
@@ -347,6 +347,135 @@ def main(conf, run, **kw) -> None:
...
+        db2 = connect(database2, threads2 or threads)
+
+    if interactive:
+        db1.enable_interactive()
+        db2.enable_interactive()
+
+    return db1, db2
+
+
+def _set_age(options, min_age, max_age, db1):
```
sungchun12, line 366:
```suggestion
def _set_age(options, min_age, max_age, db1) -> None:
```

data_diff/__main__.py
```diff
@@ -347,6 +347,135 @@ def main(conf, run, **kw) -> None:
...
+    threaded,
+    threads,
+    assume_unique_key,
+    sample_exclusive_rows,
+    materialize_all_rows,
+    table_write_limit,
+    materialize_to_table,
+    bisection_factor,
+    bisection_threshold,
+):
```
sungchun12, line 392:
```suggestion
) -> Union[JoinDiffer, HashDiffer]:
```

data_diff/__main__.py
```diff
@@ -347,6 +347,135 @@ def main(conf, run, **kw) -> None:
...
+        assert algorithm == Algorithm.HASHDIFF
+        return HashDiffer(
+            bisection_factor=bisection_factor,
+            bisection_threshold=bisection_threshold,
+            threaded=threaded,
+            max_threadpool_size=threads and threads * 2,
+        )
+
+
+def _print_result(stats, json_output, diff_iter):
```
sungchun12, line 421:
```suggestion
def _print_result(stats, json_output, diff_iter) -> None:
```

data_diff/__main__.py
sungchun12: You need to update line 9 for this type import

```python
from typing import Dict, Optional, Tuple, Union
```

sungchun12 (changes requested):
@sar009 Great job on this PR! 

Looks like you did a lot of code reorganization outside the scope of the issue you were solving for with properly closing database connections after query operations are done. 

For the changes you made in `__main__.py`, you'll need to add tests for all the private functions you created. You can create a new test file in the `/tests/` directory and name it `test_main.py` and reference the other tests to get an idea of how to test it. You can use mocking where appropriate. 

I noticed you only added the `close()` method to postgres. You'll need to define that for the other databases. Should be straightforward for other files. Example: `databricks.py`, `redshift.py`, etc. 

Let's also add a test to `test_database.py` and assert that the close method is working as expected. It'll dynamically test all the databases are closing as expected. 

Some pseudo test code.

```python
import unittest
from psycopg2 import InterfaceError

class TestDatabaseConnection(unittest.TestCase):
    def setUp(self):
        self.db = PostgreSQL(...)  # Initialize with appropriate arguments

    def tearDown(self):
        """Ensures each test starts with a fresh db connection"""
        self.db.create_connection()

    def test_close_connection(self):
        self.db.close()
        with self.assertRaises(InterfaceError):
            cursor = self.db._conn.cursor()
            cursor.execute("SELECT 1")

```

sar009:
> I noticed you only added the `close()` method to postgres. You'll need to define that for the other databases. Should be straightforward for other files. Example: `databricks.py`, `redshift.py`, etc.

All other classes are handled from the `close` method in `ThreadedDatabase` or from `Database` since `postgres.py` was overriding close and has some special way of handling connection

sar009:
> Let's also add a test to `test_database.py` and assert that the close method is working as expected. It'll dynamically test all the databases are closing as expected.

I am not sure how this will dynamically test for all of the databases. For now, I have written for Postgres

sar009:
@sungchun12 I had to remove connection sharing since it was causing test failure in parallel unit test run
https://github.com/datafold/data-diff/pull/785/files#diff-53eb316ef7b18ba47acdd955cdaa7ca5467cd5ba6f5cf8ed92c68a421902d120L90

tests/test_main.py
sungchun12, line 1: Great job overall. Added a couple tests inline.

tests/test_main.py
```diff
@@ -0,0 +1,197 @@
...
+            assert not db1._interactive
+            assert not db2._interactive
+
+        # interactive
+        db1, db2 = _get_dbs(1, db1_str, 0, db2_str, 0, True)
+        with db1, db2:
+            assert db1 != db2
+            assert db1._interactive
+            assert db2._interactive
+
```
sungchun12, line 72:
```suggestion
    def test_database_connection_failure(self) -> None:
        """Test when database connection fails."""
        with self.assertRaises(Exception):  # Assuming that connect() raises Exception on connection failure
            _get_dbs(1, "db1_str", 0, "db2_str", 0, False)

    def test_invalid_inputs(self) -> None:
        """Test invalid inputs."""
        with self.assertRaises(Exception): # Assuming that connect() raises Exception on failure
            _get_dbs(0, "", 0, "", 0, False)  # Empty connection strings

    def test_database_object(self) -> None:
        """Test returned database objects are valid and not None."""
        db1_str: str = CONN_STRINGS[db.PostgreSQL]
        db2_str: str = CONN_STRINGS[db.PostgreSQL]
        db1, db2 = _get_dbs(1, db1_str, 0, db2_str, 0, False)
        self.assertIsNotNone(db1)
        self.assertIsNotNone(db2)
        self.assertIsInstance(db1, Database)
        self.assertIsInstance(db2, Database)
        
    def test_databases_are_different(self) -> None:
        """Test separate connections for different databases."""
        db1_str: str = CONN_STRINGS[db.PostgreSQL]
        db2_str: str = CONN_STRINGS[db.MySQL]
        db1, db2 = _get_dbs(0, db1_str, 1, db2_str, 2, False)
        with db1, db2:
            self.assertIsNot(db1, db2)  # Check that db1 and db2 are not the same object
```

tests/test_main.py
```diff
@@ -0,0 +1,197 @@
...
+        _set_age(options, None, "1d", self.database)
+        assert len(options) == 1
+        assert options.get("min_update") is not None
+
+        options = {}
+        _set_age(options, "1d", "1d", self.database)
+        assert len(options) == 2
+        assert options.get("max_update") is not None
+        assert options.get("min_update") is not None
+
```
sungchun12, line 101:
```suggestion
            
    def test__set_age_db_query_failure(self):
        with self.assertRaises(Exception):
            options = {}
            _set_age(options, "1d", "1d", self.mock_database)
```

tests/test_main.py
```diff
@@ -0,0 +1,197 @@
...
+            "table1": self.table_src_name,
+            "db2": self.connection,
+            "schema2": self.dst_schema,
+            "table2": self.table_dst_name,
+        }
+        expanded_columns = _get_expanded_columns(columns, False, set(columns), **kwargs)
+
+        assert len(expanded_columns) == 3
+        assert len(set(expanded_columns) & set(columns)) == 3
+
```
sungchun12, line 160:
```suggestion

    def test__get_expanded_columns_case_sensitive(self):
        columns = ["UserID", "MovieID", "Rating"]
        kwargs = {
            "db1": self.connection,
            "schema1": self.src_schema,
            "table1": self.table_src_name,
            "db2": self.connection,
            "schema2": self.dst_schema,
            "table2": self.table_dst_name,
        }
        expanded_columns = _get_expanded_columns(columns, True, set(columns), **kwargs)

        assert len(expanded_columns) == 3
        assert len(set(expanded_columns) & set(columns)) == 3

```

tests/test_main.py
```diff
@@ -0,0 +1,197 @@
...
+        assert threaded
+        assert threads == 5
+
+        threaded, threads = _get_threads(6, 7, 8)
+        assert threaded
+        assert threads == 6
+
+        with self.assertRaises(ValueError) as value_error:
+            _get_threads(0, None, None)
+        assert str(value_error.exception) == "Error: threads must be >= 1"
```
sungchun12, line 197:
```suggestion
        assert str(value_error.exception) == "Error: threads must be >= 1"
       
        with self.assertRaises(ValueError) as value_error:
            _get_threads(-1, None, None)
        assert str(value_error.exception) == "Error: threads must be >= 1"

```

tests/test_database.py
```diff
@@ -167,3 +168,16 @@ def test_default_precision(self):
...
+    def test_close_connection_postgres(self):
+        connection: Database = get_conn(dbs.PostgreSQL)
+        with connection:
+            cursor = connection._conn.cursor()
+            cursor.execute("SELECT 1")
+
+        with self.assertRaises(InterfaceError) as interface_error:
+            cursor.execute("SELECT 1")
+
+        assert str(interface_error.exception) == "cursor already closed"
```
sungchun12, line 183:
```suggestion
@test_each_database
class TestCloseMethod(unittest.TestCase):
    def test_close_connection(self):
        connection: Database = get_conn(self.db_cls)
        
        # Perform a query to verify the connection is established
        with connection:
            cursor = connection._conn.cursor()
            cursor.execute("SELECT 1")

        # Now the connection should be closed, and trying to execute a query should fail.
        with self.assertRaises(Exception):  # Catch any type of exception.
            cursor.execute("SELECT 1")
            
```

sungchun12: This tests every database with the `@test_each_database` decorator. 

Updated `mysql.py` to get this passing. You'll have to update your branch for the below. 

```python
@attrs.define(frozen=False, init=False, kw_only=True)
class MySQL(ThreadedDatabase):
    DIALECT_CLASS: ClassVar[Type[BaseDialect]] = Dialect
    SUPPORTS_ALPHANUMS = False
    SUPPORTS_UNIQUE_CONSTAINT = True
    CONNECT_URI_HELP = "mysql://<user>:<password>@<host>/<database>"
    CONNECT_URI_PARAMS = ["database?"]

    _args: Dict[str, Any]
    _conn: Any

    def __init__(self, *, thread_count, **kw) -> None:
        super().__init__(thread_count=thread_count)
        self._args = kw

        # In MySQL schema and database are synonymous
        try:
            self.default_schema = kw["database"]
        except KeyError:
            raise ValueError("MySQL URL must specify a database")

    def create_connection(self):
        mysql = import_mysql()
        try:
            self._conn = mysql.connect(charset="utf8", use_unicode=True, **self._args)
            return self._conn
        except mysql.Error as e:
            if e.errno == mysql.errorcode.ER_ACCESS_DENIED_ERROR:
                raise ConnectError("Bad user name or password") from e
            elif e.errno == mysql.errorcode.ER_BAD_DB_ERROR:
                raise ConnectError("Database does not exist") from e
            raise ConnectError(*e.args) from e
```

tests/common.py
```diff
@@ -84,15 +85,8 @@ def get_git_revision_short_hash() -> str:
     db.MsSQL: TEST_MSSQL_CONN_STRING,
 }
 
-_database_instances = {}
-
-
-def get_conn(cls: type, shared: bool = True) -> Database:
```
sungchun12, line 90: We should be reusing this connection vs. creating a new one each time. This change should be reverted.

sungchun12: Can you point me to the failed test logs where you verified this is causing the issue?

sungchun12 (changes requested):

tests/common.py
sungchun12: Actually, I think I found it when locally testing. I'll update this thread in a bit.

tests/test_database.py
sungchun12: > 

For this close method test to pass for all databases, you'll have to update `vertica.py` and `presto.py`

based on the changes I made in this mirror branch.

sar009: @sungchun12 we do not have to close `vertica.py` since we create connections for each thread and these connections are being closed in the parent.

for `Presto` I have added the required changes.

sungchun12: @sar009 Looks like we do have to call it. I just removed the `close` method override in `vertica.py` and it failed in my mirror PR: https://github.com/datafold/data-diff/actions/runs/7534996859/job/20510263941?pr=841#step:8:75

https://github.com/datafold/data-diff/pull/841/commits/a0666de88528670d49d95782588dd4d3b1823819

The test fails because you haven't defined a ﻿`close()` method in the ﻿`Vertica` class to close the individual connection(﻿`self._conn`).

Even though you mentioned that the ﻿`close()` method should be inherited from ﻿`ThreadedDatabase`, that's not the case in this application because the ﻿`close()` method provided in ﻿`ThreadedDatabase` performs tasks that consist closing all threads and the connection in the ﻿`thread_local`. However, it will not handle closing of the individual connection (`﻿self._conn`) which is initiated in the method ﻿`create_connection()` of the ﻿`Vertica` class.

sungchun12:
@sar009 any updates?

tests/test_database.py
sungchun12: You resolved this but didn't fix this issue. Opening it back up.

sar009: @sungchun12 the `create_connection` in `vertica.py` should not set class attribute connection(`self._conn`) rather like all other classes `vertica.py` should as well return a new instance in `create_connection`, as the method `create_connection` in `ThreadedDatabase` is called from `set_conn` in `ThreadedDatabase` and the db connection is managed in `self.thread_local` of `ThreadedDatabase`. The connection set in this fashion is closed from close method of `ThreadedDatabase`.

Your pr is failing after commenting the close function in `vertica.py` because the connection is used from `self._conn` and not from `self.thread_local.conn` of `ThreadedDatabase`. If you revert your change in `create_connection` of `vertica.py` to original as follow the build should be successful.
```python
    def create_connection(self):
        vertica = import_vertica()
        try:
            c = vertica.connect(**self._args)
            return c
        except vertica.errors.ConnectionError as e:
            raise ConnectError(*e.args) from e
```

P.S. only Postgres' behavior is different I will take a look at that as well once this PR is merged.

sungchun12: @sar009 I reverted to the original and it fails the test both with and without the `_conn` attribute added. 

https://github.com/datafold/data-diff/actions/runs/7587669203/job/20668531425?pr=841
https://github.com/datafold/data-diff/actions/runs/7587791525/job/20668911798?pr=841

Unless there's something else I'm missing, I recommend you follow my code suggestion!

After you make the update, let me know when this is ready for a full review. We're almost at the finish line!

sar009: @sungchun12 Thanks You, the tests are failing because additionally what I have posted above the test method in `test_database.py` should look something like this

```
@test_closeable_databases
class TestCloseMethod(unittest.TestCase):
    def test_close_connection(self):
        database: Database = get_conn(self.db_cls)

        # Perform a query to verify the connection is established
        with database:
            database.query("SELECT 1")

        # Now the connection should be closed, and trying to execute a query should fail.
        with self.assertRaises(Exception):  # Catch any type of exception.
            database.query("SELECT 1")
```

in your code you are accessing the `_conn` object explicitly. Let the db class decide how to manage the connection. Also as with all the code I am recommending is already in this PR and the tests are executing fine.

sungchun12 (approved):

tests/test_database.py
sar009: Thanks for merging the PR @sungchun12

PR #792, opened 2023-12-06
Accept either key file path or file itself in Snowflake

vvkh:


data_diff/databases/snowflake.py
```diff
@@ -173,19 +173,27 @@ def __init__(self, *, schema: str, **kw):
 
         assert '"' not in schema, "Schema name should not contain quotes!"
         # If a private key is used, read it from the specified path and pass it as "private_key" to the connector.
-        if "key" in kw:
```
anatoly-scherbakov, line 176: I wonder if we could change the function signature to this:

```python
def __init__(self, *, schema: str, key: str | None, key_file_path: str | None, **kw):
```

which would be a step towards normal typing in this code.

nolar (approved):

anatoly-scherbakov (approved):

anatoly-scherbakov (approved):

PR #794, opened 2023-12-06
Revert "mssql normalize_uuid" and fix the overall problem

dlawin:
normalize_uuid change was just a symptom of problem with mssql's limit_select

- revert normalize_uuid change
- fix limit_select for mssql

### DIFF
--- a/data_diff/databases/mssql.py
+++ b/data_diff/databases/mssql.py
@@ -113,9 +113,9 @@ class Dialect(BaseDialect):
     def limit_select(
         self,
         select_query: str,
-        offset: int | None = None,
-        limit: int | None = None,
-        has_order_by: bool | None = None,
+        offset: Optional[int] = None,
+        limit: Optional[int] = None,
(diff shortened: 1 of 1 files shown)


pik94 (approved):

pik94:
@dlawin i approved it just in case not to block if it is a hot fix. But for me it is unclear how uuid normalization breaks it. Could you clarify?

data_diff/databases/mssql.py
```diff
@@ -155,9 +156,6 @@ def md5_as_int(self, s: str) -> str:
     def md5_as_hex(self, s: str) -> str:
         return f"HashBytes('MD5', {s})"
 
-    def normalize_uuid(self, value: str, coltype: ColType_UUID) -> str:
```
pik94, line 158: What was a problem with this function?

dlawin: this function was a red herring, the actual issue was with limited_select

PR #804, opened 2023-12-18
Retrieve collations from the schema (and refactor the column info structures)

nolar:
_Requires #813 because of a minor merge conflict; otherwise independent._


nolar:
Supersedes by #814 due to not running CI here for some reason.

PR #805, opened 2023-12-18
yield list vs. indiv tuples

dlawin:
Add an optional parameter `yield_list` to differ

Setting `yield_list` to True allows the diffs to be consumed as a list -- which is how the diffs are actually generated. That list represents the diffs found over a key range.

### DIFF
--- a/data_diff/diff_tables.py
+++ b/data_diff/diff_tables.py
@@ -6,7 +6,7 @@ from abc import ABC, abstractmethod
 from enum import Enum
 from contextlib import contextmanager
 from operator import methodcaller
-from typing import Dict, Set, Tuple, Iterator, Optional
+from typing import Dict, Set, List, Tuple, Iterator, Optional
 from concurrent.futures import ThreadPoolExecutor, as_completed
 
 import attrs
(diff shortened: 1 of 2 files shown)


data_diff/thread_utils.py
```diff
@@ -56,19 +55,24 @@ class ThreadedYielder(Iterable):
...
         self._exception = None
+        self.yield_list = yield_list
 
     def _worker(self, fn, *args, **kwargs):
         try:
             res = fn(*args, **kwargs)
             if res is not None:
-                self._yield += res
+                if self.yield_list:
+                    self._yield.append(res)
```
dlawin, line 74: e.g. when true, add the list[diff] to the _yield deque -> deque[list[diff]]

Instead of the previous deque[diff]

data_diff/diff_tables.py
```diff
@@ -300,9 +304,9 @@ def _bisect_and_diff_tables(self, table1: TableSegment, table2: TableSegment, in
             f"size: table1 <= {btable1.approximate_size()}, table2 <= {btable2.approximate_size()}"
         )
 
-        ti = ThreadedYielder(self.max_threadpool_size)
+        ti = ThreadedYielder(self.max_threadpool_size, self.yield_list)
         # Bisect (split) the table into segments, and diff them recursively.
-        ti.submit(self._bisect_and_diff_segments, ti, btable1, btable2, info_tree)
+        ti.submit(self._bisect_and_diff_segments_root, ti, btable1, btable2, info_tree, priority=999)
```
dlawin, line 309: `priority=999` want to call this out

nolar (approved):

PR #806, opened 2023-12-19
Resolve failing mssql tests

pik94:
This PR affects MSSQL connector
- Fix timestamp normalization
- Fix number normalization
- Partly fix uuid normalization: text UUID are supported, native UUID is turned of

The reason why i decided to exclude native UUID type from the current implementation is casting.
MSSQL UUID values are stored in upper case, whereas other databases generally use lower case. Moreover, when casting MSSQL UUIDs to python UUIDs, we lose upper case, and it leads to incorrect sql query gereration. 
For example, we want to compare MSSQL vs PostgreSQL tables with UUID primary keys. For MSSQL a query which calculates md5 hash looks like:
```
SELECT 
    COUNT(*), 
    MD5(... [col] ...) 
FROM UUID_TABLE 
WHERE ([col] >= 'af8ef74d-9b83-11ee-8e2a-000000000000') AND ([col] < 'af8efd8d-f0d8-6743-ef3c-000000000011')
```
UUID values in `WHERE` clause are in lower case after query building, whereas we need upper case for them, and as the result, we have an empty sample in contrast to PostgreSQL.

I have not managed to fix this on the MSSQL connection sidbe, it should be done in the core side.

### DIFF
--- a/data_diff/databases/mssql.py
+++ b/data_diff/databases/mssql.py
@@ -69,7 +69,7 @@ class Dialect(BaseDialect):
         "varbinary": Text,
         "xml": Text,
         # UUID
-        "uniqueidentifier": Native_UUID,
+        # "uniqueidentifier": Native_UUID,  # It is not supported yet
         # Bool
         "bit": Boolean,
         # JSON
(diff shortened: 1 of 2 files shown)


nolar:
Related:
* Varchar(max) https://github.com/datafold/data-diff/pull/825
* UUID case preservation https://github.com/datafold/data-diff/pull/825

PR #809, opened 2023-12-20
Better Action Next Step when using CLI

sungchun12:
Point users to an evolved CLI experience by trialing Datafold Cloud. 

image

@KiraFuruichi @leoebfolsom  Included you to verify that's a good entrypoint URL and language call out. If you approve, I can simply squash and merge as this code change is cosmetic.

### DIFF
--- a/data_diff/dbt.py
+++ b/data_diff/dbt.py
@@ -50,7 +50,7 @@ from data_diff.utils import (
 
 logger = getLogger(__name__)
 CLOUD_DOC_URL = "https://docs.datafold.com/development_testing/cloud"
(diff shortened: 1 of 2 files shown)


data_diff/dbt.py
```diff
@@ -558,6 +560,8 @@ def _email_signup() -> None:
 
 def _extension_notification() -> None:
     if bool_notify_about_extension():
```
dlawin, line 562: I think we should rename these, but not blocking

dlawin (approved):

data_diff/dbt.py
sungchun12: Cool. I'll merge this then!

leoebfolsom (commented):
👍

PR #812, opened 2023-12-26
[to#811]Fix special characters in PG url and Mysql connection reconnect

### LINKED ISSUE #811: PG connection URL has special characters in password
**Describe the bug**
When connect to PG, if my password has special characters, may be like '@'.  The code goes to dns parser will fail to extract the right password.

idling11:
resolves #811 

1. fix special characters in PG url password
2. fix mysql connection not available after a long idle time

### DIFF
--- a/data_diff/databases/mysql.py
+++ b/data_diff/databases/mysql.py
@@ -19,7 +19,8 @@ from data_diff.databases.base import (
     ThreadedDatabase,
     import_helper,
     ConnectError,
-    BaseDialect, ThreadLocalInterpreter,
+    BaseDialect, 
+    ThreadLocalInterpreter,
 )
 from data_diff.databases.base import (
     MD5_HEXDIGITS,
--- a/data_diff/databases/mysql.py
+++ b/data_diff/databases/mysql.py
@@ -157,3 +157,4 @@ class MySQL(ThreadedDatabase):
         if not self.thread_local.conn.is_connected():
             self.thread_local.conn.ping(reconnect=True, attempts=3, delay=5)
         return self._query_conn(self.thread_local.conn, sql_code)
+       
--- a/data_diff/databases/mysql.py
+++ b/data_diff/databases/mysql.py
@@ -1,4 +1,4 @@
-from typing import Any, ClassVar, Dict, Type
+from typing import Any, ClassVar, Dict, Type, Union
 
 import attrs
 
@@ -19,7 +19,7 @@ from data_diff.databases.base import (
     ThreadedDatabase,
     import_helper,
     ConnectError,
-    BaseDialect,
+    BaseDialect, ThreadLocalInterpreter,
 )
 from data_diff.databases.base import (
     MD5_HEXDIGITS,
@@ -148,3 +148,11 @@ class MySQL(ThreadedDatabase):
             elif e.errno == mysql.errorcode.ER_BAD_DB_ERROR:
                 raise ConnectError("Database does not exist") from e
             raise ConnectError(*e.args) from e
+
+    def _query_in_worker(self, sql_code: Union[str, ThreadLocalInterpreter]):
+        "This method runs in a worker thread"
+        if self._init_error:
+            raise self._init_error
+        if not self.thread_local.conn.is_connected():
+            self.thread_local.conn.ping(reconnect=True, attempts=3, delay=5)
+        return self._query_conn(self.thread_local.conn, sql_code)
--- a/data_diff/databases/mysql.py
+++ b/data_diff/databases/mysql.py
@@ -19,7 +19,7 @@ from data_diff.databases.base import (
     ThreadedDatabase,
     import_helper,
     ConnectError,
-    BaseDialect, 
+    BaseDialect,
     ThreadLocalInterpreter,
 )
 from data_diff.databases.base import (
@@ -157,4 +157,3 @@ class MySQL(ThreadedDatabase):
         if not self.thread_local.conn.is_connected():
             self.thread_local.conn.ping(reconnect=True, attempts=3, delay=5)
         return self._query_conn(self.thread_local.conn, sql_code)
-       
--- a/data_diff/databases/postgresql.py
+++ b/data_diff/databases/postgresql.py
@@ -1,5 +1,5 @@
 from typing import Any, ClassVar, Dict, List, Type
-
+from urllib.parse import unquote
 import attrs
 
 from data_diff.abcs.database_types import (
@@ -168,6 +168,7 @@ class PostgreSQL(ThreadedDatabase):
 
         pg = import_postgresql()
         try:
+            self._args["password"] = unquote(self._args["password"])
             self._conn = pg.connect(
                 **self._args, keepalives=1, keepalives_idle=5, keepalives_interval=2, keepalives_count=2
             )
--- a/tests/test_postgresql.py
+++ b/tests/test_postgresql.py
@@ -153,5 +153,3 @@ class TestSpecialCharacterPassword(unittest.TestCase):
 
         connection_verified = connect(db_url)
         assert connection_verified._args.get("password") == password
-
-
--- a/tests/test_postgresql.py
+++ b/tests/test_postgresql.py
@@ -153,3 +153,4 @@ class TestSpecialCharacterPassword(unittest.TestCase):
 
         connection_verified = connect(db_url)
         assert connection_verified._args.get('password') == password
+
--- a/tests/test_postgresql.py
+++ b/tests/test_postgresql.py
@@ -1,9 +1,11 @@
 import unittest
 
+from urllib.parse import quote
 from data_diff.queries.api import table, commit
 from data_diff import TableSegment, HashDiffer
 from data_diff import databases as db
-from tests.common import get_conn, random_table_suffix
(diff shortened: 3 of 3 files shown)


data_diff/databases/postgresql.py
```diff
@@ -168,6 +168,7 @@ def create_connection(self):
 
         pg = import_postgresql()
         try:
+            self._args["password"] = unquote(self._args["password"])
```
sungchun12, line 171: How are you passing in this parameter when connecting? I'm able to use a password with special characters just fine when connecting using the `connect_to_table` method with a dictionary. This test case passes when I run it without your changes. I recommend use this method instead of constructing the URL connection string. Let me know if I'm missing something! 

```python
class TestSpecialCharacterPassword(unittest.TestCase):
    def setUp(self) -> None:
        self.connection = get_conn(db.PostgreSQL)

        table_suffix = random_table_suffix()

        self.table_name = f"table{table_suffix}"
        self.table = table(self.table_name)

    def test_special_char_password(self):
        password = "passw!!!@rd"
        # Setup user with special character '@' in password
        self.connection.query("DROP USER IF EXISTS test;", None)
        self.connection.query(f"CREATE USER test WITH PASSWORD '{password}';", None)
        
        connect_to_table(
            {
                "driver": "postgresql",
                "host": "localhost",
                "port": 5432,
                "dbname": "postgres",
                "user": "test",
                "password": password,
            },
            self.table_name,
        )
```

data_diff/databases/mysql.py
```diff
@@ -148,3 +148,11 @@ def create_connection(self):
             elif e.errno == mysql.errorcode.ER_BAD_DB_ERROR:
                 raise ConnectError("Database does not exist") from e
             raise ConnectError(*e.args) from e
+
+    def _query_in_worker(self, sql_code: Union[str, ThreadLocalInterpreter]):
```
sungchun12, line 152: You defined this private method here, but where are you calling it?

sungchun12 (changes requested):

data_diff/databases/postgresql.py
idling11: Thx for replay. As you said, If you pass a dict to `connect_to_table` function, there is nothing wrong. But when passing a url with special characters , connection will failed because of dns wrong parsing. I think that since we allow `connect_to_table` to pass in the url string as a parameter, we must also deal with the possible consequences of url errors.

idling11: for example, If i write codes like this: `db_url = f"postgresql://{config['user']}:{config['password']}@{config['host']}:{config['port']}/{config['database']}"`. And these parameters are from out conf files,  when the `config['password']` has special characters like `@`, dns parse will failed.

data_diff/databases/mysql.py
idling11: When `Database` calling `query` function,  `_query` func in `ThreadedDatabase` class will be called. Then `_query_in_worker` in `ThreadedDatabase` will be executed. I put this func here to execute `Mysql`.`_query_in_worker` instead of `_query_in_worker` in `ThreadedDatabase`. If there is any problem, please let me know, thank you

data_diff/databases/postgresql.py
sungchun12: Thanks for providing the example url construction code. Fair point, and verified your fix works. Can you add the below unit test to your branch? 

```python
class TestSpecialCharacterPassword(unittest.TestCase):
    def setUp(self) -> None:
        self.connection = get_conn(db.PostgreSQL)

        table_suffix = random_table_suffix()

        self.table_name = f"table{table_suffix}"
        self.table = table(self.table_name)

    def test_special_char_password(self):
        password = "passw!!!@rd"
        # Setup user with special character '@' in password
        self.connection.query("DROP USER IF EXISTS test;", None)
        self.connection.query(f"CREATE USER test WITH PASSWORD '{password}';", None)

        password_quoted = quote(password)
        db_config = {
                "driver": "postgresql",
                "host": "localhost",
                "port": 5432,
                "dbname": "postgres",
                "user": "test",
                "password": password_quoted,
            }

        # verify pythonic connection method
        connect_to_table(
            db_config,
            self.table_name,
        )

        # verify connection method with URL string unquoted after it's verified
        db_url = f"postgresql://{db_config['user']}:{db_config['password']}@{db_config['host']}:{db_config['port']}/{db_config['dbname']}"

        connection_verified = connect(db_url)
        assert connection_verified._args.get('password') == password
```

data_diff/databases/mysql.py
sungchun12: Got it, thanks for explaining that and realizing that it does override the inherited class function with the MySQL-specific implementation.

sungchun12:
Working on a fix for the ruff formatter.

data_diff/databases/postgresql.py
sungchun12: Here are the updated imports.

```python
import unittest

from urllib.parse import quote
from data_diff.queries.api import table, commit
from data_diff import TableSegment, HashDiffer
from data_diff import databases as db
from tests.common import get_conn, random_table_suffix, connect
from data_diff import connect_to_table
```

idling11: Already add UT in test_postgresql.py. Thanks for providing the codes.

data_diff/databases/mysql.py
```diff
@@ -19,7 +19,7 @@
     ThreadedDatabase,
     import_helper,
     ConnectError,
-    BaseDialect,
+    BaseDialect, ThreadLocalInterpreter,
```
sungchun12, line 22:
```suggestion
    BaseDialect, 
    ThreadLocalInterpreter,
```

sungchun12 (dismissed):

data_diff/databases/mysql.py
```diff
@@ -148,3 +149,11 @@ def create_connection(self):
...
                 raise ConnectError("Database does not exist") from e
             raise ConnectError(*e.args) from e
+
+    def _query_in_worker(self, sql_code: Union[str, ThreadLocalInterpreter]):
+        "This method runs in a worker thread"
+        if self._init_error:
+            raise self._init_error
+        if not self.thread_local.conn.is_connected():
+            self.thread_local.conn.ping(reconnect=True, attempts=3, delay=5)
+        return self._query_conn(self.thread_local.conn, sql_code)
```
sungchun12, line 159:
```suggestion
        return self._query_conn(self.thread_local.conn, sql_code)
       
```

tests/test_postgresql.py
```diff
@@ -113,3 +115,41 @@ def test_100_fields(self):
...
+        connect_to_table(
+            db_config,
+            self.table_name,
+        )
+
+        # verify connection method with URL string unquoted after it's verified
+        db_url = f"postgresql://{db_config['user']}:{db_config['password']}@{db_config['host']}:{db_config['port']}/{db_config['dbname']}"
+
+        connection_verified = connect(db_url)
+        assert connection_verified._args.get('password') == password
```
sungchun12, line 155:
```suggestion
        assert connection_verified._args.get('password') == password

```

data_diff/databases/mysql.py
```diff
@@ -19,7 +19,8 @@
     ThreadedDatabase,
     import_helper,
     ConnectError,
-    BaseDialect,
+    BaseDialect,
```
sungchun12, line 22:
```suggestion
    BaseDialect,
```

data_diff/databases/mysql.py
```diff
@@ -148,3 +149,12 @@ def create_connection(self):
...
             raise ConnectError(*e.args) from e
+
+    def _query_in_worker(self, sql_code: Union[str, ThreadLocalInterpreter]):
+        "This method runs in a worker thread"
+        if self._init_error:
+            raise self._init_error
+        if not self.thread_local.conn.is_connected():
+            self.thread_local.conn.ping(reconnect=True, attempts=3, delay=5)
+        return self._query_conn(self.thread_local.conn, sql_code)
+
```
sungchun12, line 160:
```suggestion

```

tests/test_postgresql.py
```diff
@@ -113,3 +115,42 @@ def test_100_fields(self):
...
```
sungchun12, line 155:
```suggestion
        assert connection_verified._args.get("password") == password

```

sungchun12:
Validate all unit tests pass: https://github.com/datafold/data-diff/pull/819#issue-2058648932

sungchun12:
@idling11 Can you run `ruff format .` on your forked branch and commit the change? Once that formatter test passes, I'll squash and merge this and we can cut a patch release!

idling11:
@sungchun12 Formated files and submitted.

sungchun12 (approved):
All tests pass in a native branch for extra due diligence: https://github.com/datafold/data-diff/pull/819

sungchun12:
@idling11 Just merged and we'll likely cut a patch release next week! In the short term, you can use the datafold master branch!

sungchun12:
Thank you for contributing and your patience in this process. You helped uncover a lot of issues when work with forked branches in our CI process. Next time you contribute, you'll have a much better experience and it's thanks to you :)

PR #814, opened 2023-12-27
Retrieve collations from the schema (and refactor the column info structures)

nolar:
~_Requires #813 because of a minor merge conflict; otherwise independent._~

_A replay for #804, which did not run CI for some reason._


dagadbm (approved):
to unblock if needed but needs proper review

PR #823, opened 2023-12-29
Fix precision recognition

vvkh:
In Postgres, if type is specified as `NUMERIC` (no scale/precision parameters), the effective scale is 16383 ([docs](https://www.postgresql.org/docs/current/datatype-numeric.html#DATATYPE-NUMERIC-TABLE)).
In BigQuery, the effective scale in this case is 9 ([community forum](https://www.googlecloudcommunity.com/gc/Databases/What-s-default-Scale-and-Precision-of-Numeric-type-in-BigQuery/m-p/612292#M1281)).

### DIFF
--- a/data_diff/databases/bigquery.py
+++ b/data_diff/databases/bigquery.py
@@ -75,6 +75,9 @@ class Dialect(BaseDialect):
     }
     TYPE_ARRAY_RE = re.compile(r"ARRAY<(.+)>")
     TYPE_STRUCT_RE = re.compile(r"STRUCT<(.+)>")
+    # https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#parameterized_decimal_type
+    # The default scale is 9, which means a number can have up to 9 digits after the decimal point.
+    DEFAULT_NUMERIC_PRECISION = 9
 
     def random(self) -> str:
         return "RAND()"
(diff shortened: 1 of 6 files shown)


data_diff/databases/bigquery.py
```diff
@@ -76,6 +76,11 @@ class Dialect(BaseDialect):
     }
     TYPE_ARRAY_RE = re.compile(r"ARRAY<(.+)>")
     TYPE_STRUCT_RE = re.compile(r"STRUCT<(.+)>")
+    # [BIG]NUMERIC, [BIG]NUMERIC(precision, scale), [BIG]NUMERIC(precision)
+    TYPE_NUMERIC_RE = re.compile(r"^((BIG)?NUMERIC)(?:\((\d+)(?:, (\d+))?\))?$")
```
dlawin, line 80: [BIG]NUMERIC() not possible?

dlawin (approved):

data_diff/databases/bigquery.py
vvkh: Yep, it's invalid in BQ.
image

nolar, line 80: Is it guaranteed to have strictly 1 space after the comma? Never 0?

vvkh: Yes, the space is always there even if it user didn't add it explicitly.

Issue #833, opened 2024-01-03
Ability to install all database adapters


sar009:
**Is your feature request related to a problem? Please describe.**
I am using almost all databases and I need to install the adapter for each database as
```shell
pip install data-diff 'data-diff[postgresql,snowflake ,redshift,mysql,presto,oracle]' -U
```
rather if I could use something like the below and install all database adapters
```shell
pip install data-diff 'data-diff[with_everything]' -U
```

**Describe the solution you'd like**
add `with_everything` option

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

sar009:
fixed by https://github.com/datafold/data-diff/pull/834

PR #834, opened 2024-01-03
Ability to install all database adapters.

sar009:
use 
```
pip install data-diff 'data-diff[with_everything]' -U
```
to install all database adapters

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,79 +9,61 @@ data-diff: Compare datasets fast, within or across SQL databases
 </h2>
 <br>
 
-> [Make sure to join us at our virtual hands-on lab series where our team walks through live how to get set-up with it!](https://www.datafold.com/virtual-hands-on-lab)
+> [Join our live virtual lab series to learn how to set it up!](https://www.datafold.com/virtual-hands-on-lab)
+
+# What's a Data Diff?
+A data diff is the value-level comparison between two tables—used to identify critical changes to your data and guarantee data quality.
+
+There is a lot you can do with data-diff: you can test SQL code by comparing development or staging environment data to production, or compare source and target data to identify discrepancies when moving data between databases.
 
 # Use Cases
 
-## Data Migration & Replication Testing
-Compare source to target and check for discrepancies when moving data between systems:
-- Migrating to a new data warehouse (e.g., Oracle > Snowflake)
-- Converting SQL to a new transformation framework (e.g., stored procedures > dbt)
-- Continuously replicating data from an OLTP DB to OLAP DWH (e.g., MySQL > Redshift)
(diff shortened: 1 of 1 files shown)


sar009:
fixes https://github.com/datafold/data-diff/issues/833

pyproject.toml
```diff
@@ -42,7 +42,7 @@ preql = {version="^0.2.19", optional=true}
 vertica-python = {version="*", optional=true}
 urllib3 = "<2"
 oracledb = {version = "*", optional=true}
-pyodbc = {version="^4.0.39", optional=true}
+pyodbc = {version=">=4.0.39", optional=true}
```
sungchun12, line 45:
```suggestion
pyodbc = {version="^4.0.39", optional=true}
```
We only install minor patches to prevent breaking changes. If you have a compelling reason to update this, open a separate PR.

pyproject.toml
```diff
@@ -80,6 +80,10 @@ trino = ["trino"]
 clickhouse = ["clickhouse-driver"]
 vertica = ["vertica-python"]
 duckdb = ["duckdb"]
+with_everything = [
```
sungchun12, line 83:
```suggestion
all-dbs = [
```

README.md
```diff
@@ -81,6 +81,11 @@ More information about the algorithm and performance considerations can be found
 pip install data-diff 'data-diff[postgresql,snowflake]' -U
 ```
 
+Additionally, you can install all database adapters as follows.
+```
+pip install data-diff 'data-diff[with_everything]' -U
```
sungchun12, line 86:
```suggestion
pip install data-diff 'data-diff[all-dbs]' -U
```

sungchun12 (changes requested):
Made some suggestions. Overall, looks good!

pyproject.toml
sar009: The dependency is related to MSSQL and MSSQL is not supported natively. That's why I decided to update the same PR. The right thing to do would be to remove all MSSQL-related files.

sungchun12: We support MSSQL in Datafold Cloud so some dependencies are inherited from this repo. Once you commit the suggestion, I'll approve and merge!

pyproject.toml
```diff
@@ -80,6 +80,10 @@ trino = ["trino"]
 clickhouse = ["clickhouse-driver"]
 vertica = ["vertica-python"]
 duckdb = ["duckdb"]
+all-dbs = [
```
sungchun12, line 83: Can you comment out this line? This makes it clear why this isn't part of the `all-dbs` extra list. 

```poetry
mssql = ["pyodbc"]
```

PR #843, opened 2024-01-08
remove local computer piece

sungchun12:
This can be run in simple scripts. Let's not confuse people to think it only runs on your local computer


elliotgunn (approved):
Looks good!

PR #842, opened 2024-01-08
Ability to install all database adapters.

sar009:
use

```
pip install data-diff 'data-diff[all-local-dbs]' -U
```

to install all database adapters in local

PR #845, opened 2024-01-09
Microsoft SQL Server Support Docs

sungchun12:
Update docs to reflect that MS SQL is supported by data-diff


sungchun12:
Fixes the mistake I made when reviewing #842

dlawin (approved):

PR #846, opened 2024-01-09
fix for Warning: poetry.lock is not consistent with pyproject.toml

dlawin:
Noticed this locally:
```
Warning: poetry.lock is not consistent with pyproject.toml. You may be getting improper dependencies. Run `poetry lock [--no-update]` to fix it.
```


sungchun12:
Ahh good catch. I just ran this in my PR, let's close this out and use that for the poetry lock update. #845

PR #857, opened 2024-01-24
Evolve dbt data diff output

### LINKED ISSUE #409: Make all format options available when using `--dbt`

sungchun12:
resolves #409

Cooking up progress
> note: may be different from initial design in referenced issue as development progresses

`data-diff --dbt`
image

`data-diff --dbt --cloud`
> Note: Columns added [+1]: {'NEW_COLUMN'} does NOT appear in this output because the API results payload doesn't include that column at all in the diff. This is intended behavior.
image

Adding a `--show` sample diff command is NOT worth implementing until the [Datafold REST API](https://docs.datafold.com/reference/cloud/rest/get-diff-results-api-v-1-datadiffs-datadiff-id-results-get) is updated to return a limited a result set. If implemented today, it would slow down `data-diff` significantly and consume more memory given 1,000,000 results can be returned by the API with no pre-filtering possible. 

**Datafold API Needs**
- [Limit diff results on the Datafold backend](https://linear.app/datafold/issue/CX-2329/limit-diff-results-returned-in-the-config-submissions)
- [Get the REST API to return table object dependencies as well as data apps](https://linear.app/datafold/issue/CX-2330/get-the-rest-api-to-return-table-object-dependencies-as-well-as-data)


data_diff/dbt.py
```diff
@@ -455,32 +466,57 @@ def _cloud_diff(
         rows_removed_count = diff_results.pks.exclusives[0]
 
         rows_updated = diff_results.values.rows_with_differences
-        total_rows = diff_results.values.total_rows
-        rows_unchanged = int(total_rows) - int(rows_updated)
+        total_rows_table1 = diff_results.pks.total_rows[0]
+        total_rows_table2 = diff_results.pks.total_rows[1]
+        total_rows_diff = total_rows_table2 - total_rows_table1
+
+        rows_unchanged = int(total_rows_table1) - int(rows_updated) - int(rows_removed_count)
```
sungchun12, line 473: Updated this math as it wasn't correct before

dlawin (approved):

Issue #859, opened 2024-01-25
Namespacing CLI variables in __main__


sar009:
**Is your feature request related to a problem? Please describe.**
No, `__main__` is bloated, one of the reason being, the cli variables are used without namespacing

**Describe the solution you'd like**
Namespace cli variable

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #860, opened 2024-01-25
Namespacing CLI options

sar009:
The `__main__.py` is blotted with a lot of CLI variables impacting readability and loose typing. So namespacing the CLI options under the class `CliOptions`.


sar009:
fixing https://github.com/datafold/data-diff/issues/859

sungchun12:
Can you add a fuller description on your approach in this PR? I'll plan to review this next week when it's ready!

sar009:
@sungchun12 I have updated the description please have a look.

data_diff/__init__.py
```diff
@@ -14,7 +14,7 @@
 def connect_to_table(
     db_info: Union[str, dict],
     table_name: Union[DbPath, str],
-    key_columns: str = ("id",),
+    key_columns: Tuple[str] = ("id",),
```
sungchun12, line 17: thanks, good catch. You're reminding me that we need to add static type checking to this codebase

data_diff/config.py
```diff
@@ -36,15 +37,17 @@ def _apply_config(config: Dict[str, Any], run_name: str, kw: Dict[str, Any]):
     else:
         run_name = "default"
 
-    if kw.get("database1") is not None:
+    # if db details are provided from cli that has precedence
+    if cli_options.database1 is not None:
         for attr in ("table1", "database2", "table2"):
-            if kw[attr] is None:
+            if cli_options.__getattribute__(attr) is None:
                 raise ValueError(f"Specified database1 but not {attr}. Must specify all 4 arguments, or neither.")
```
sungchun12, line 44:
```suggestion
                raise ValueError(f"Specified database1 but not {attr}. Must specify all 4 arguments (database1, table1, database2, table2), or none.")
```

Make it clear what 4 arguments are needed

data_diff/cli_options.py
```diff
@@ -0,0 +1,48 @@
+from dataclasses import dataclass
+from typing import Optional, Literal, Dict, Union, Tuple
+
+
+@dataclass
+class CliOptions:
```
sungchun12, line 6: Seeing all this really puts into perspective just how many CLI options are available that need validation. Can you refactor this using pydantic? 

Until we have static type checking in place, runtime type validation will serve this better

quick pseudo-code
```python
from typing import Optional, Literal, Dict, Union, Tuple
from pydantic import BaseModel


class CliOptions(BaseModel):
    bisection_factor: int
    bisection_threshold: int
    table_write_limit: int
    database1: Union[str, Dict, None] = None
    table1: Optional[str] = None
    database2: Union[str, Dict, None] = None
    table2: Optional[str] = None
    key_columns: Tuple[str, ...] = ()
    update_column: Optional[str] = None
    columns: Tuple[str, ...] = ()
    limit: Optional[int] = None
    materialize_to_table: Optional[str] = None
    min_age: Optional[str] = None
    max_age: Optional[str] = None
    stats: bool = False
    debug: bool = False
    json_output: bool = False
    verbose: bool = False
    version: bool = False
    interactive: bool = False
    no_tracking: bool = False
    case_sensitive: bool = False
    assume_unique_key: bool = False
    sample_exclusive_rows: bool = False
    materialize_all_rows: bool = False
    threads: Union[None, int, Literal["serial"]] = None
    threads1: Optional[int] = None
    threads2: Optional[int] = None
    threaded: bool = False
    where: Optional[str] = None
    algorithm: Literal["auto", "joindiff", "hashdiff"] = "auto"
    conf: Optional[str] = None
    run: Optional[str] = None
    dbt: bool = False
    cloud: bool = False
    dbt_profiles_dir: Optional[str] = None
    dbt_project_dir: Optional[str] = None
    select: Optional[str] = None
    state: Optional[str] = None
    prod_database: Optional[str] = None
    prod_schema: Optional[str] = None
    __conf__: Optional[Dict] = None
```

sungchun12: You will have to change the attribute calls for the rest of the files as a result

tests/test_main.py
```diff
@@ -209,41 +256,64 @@ def test__get_expanded_columns_case_sensitive(self):
...
-
-        threaded, threads = _get_threads(5, None, None)
-        assert threaded
-        assert threads == 5
-
-        threaded, threads = _get_threads(6, 7, 8)
-        assert threaded
-        assert threads == 6
+            _set_threads(cli_options)
+        assert str(value_error.exception) == "Error: threads must be of type int, or value must be 'serial'."
```
sungchun12, line 290: pydantic will resolve this at runtime validation, but this custom error message is nice

sungchun12 (changes requested):
Overall, I really like this approach and you did a great job of cleaning up the implementation emails and reducing `kwargs` bloat for function calls.

Once you update the data class for a pydantic class and update the relevant attribute calls, I'll give this another review and aim to merge this week!

sar009:
@sungchun12 I have addressed the review comments please take a look

data_diff/__main__.py
```diff
@@ -162,50 +181,67 @@ def write_usage(self, prog: str, args: str = "", prefix: Optional[str] = None) -
...
     metavar="AGE",
+    type=str,
 )
 @click.option(
-    "--max-age", default=None, help="Considers only rows younger than specified. See --min-age.", metavar="AGE"
+    "--max-age",
+    default=None,
+    help="Considers only rows younger than specified. See --min-age.",
+    metavar="AGE",
+    type=str,
```
sungchun12, line 191: why is this a string?

sungchun12:
Mirror this PR for the full test suite to uncover any edge cases: https://github.com/datafold/data-diff/pull/871

data_diff/cli_options.py
```diff
@@ -0,0 +1,48 @@
+from typing import Optional, Literal, Dict, Union, Tuple
+
+from pydantic import BaseModel, PositiveInt
+
+
+class CliOptions(BaseModel):
```
sungchun12, line 6: This class is bloated and will become bloated over time, but I'm willing to accept that tradeoff for better runtime validation of flag types and simplification of `kwargs**` across method calls. Overall, this PR has opened my eyes that a major refactor needs to happen across lots of the codebase for simpler data flows, but in the short-term, this approach is worth it.

sungchun12 (changes requested):
Overall, really good!

Just need new tests for these methods/functions given this codebase hasn't tested them before and we need to make sure the new args we're passing are doing exactly as expected.

- `_apply_config`
- `apply_config_from_file`
- `_print_result`
- `_data_diff`
- `create_schema`
- `diff_schemas`

data_diff/__main__.py
```diff
@@ -162,50 +181,67 @@ def write_usage(self, prog: str, args: str = "", prefix: Optional[str] = None) -
...
+        f"Maximum number of rows to write when creating materialized or sample tables, per thread. "
+        f"Default={TABLE_WRITE_LIMIT}"
+    ),
+    type=int,
     metavar="COUNT",
 )
 @click.option(
     "-j",
     "--threads",
     default=None,
```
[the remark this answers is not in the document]
sar009: https://github.com/datafold/data-diff/pull/860/files#diff-c4d05fd628acd83c42409afcbd38e6fca96261992bea7d774dd5d7ab598716a2R498

threads can be `serial`

Issue #885, opened 2024-04-16
[SQL Server] TRIM() is not available in versions of sql server prior to 17


MiConnell:
**Describe the bug**
data diff runs a query with `TRIM()` but sql server <17 does not support this.

Make sure to include the following (minus sensitive information):
command:

```
data-diff -d  <redshift_uri>   <schema>.<table> <mssql_uri>   dbo.<TABLE>   -k <KEY_COLUMN> 
```

output:
```
        DEBUG    Available mutual columns: set()                                                                                                              __main__.py:578
         INFO     Diffing using columns: key=('COLUMN',) update=None extra=().                                                                   __main__.py:598
         DEBUG    Running SQL (Redshift): ('schema', 'table')                                                                                       base.py:980
                  SELECT * FROM (SELECT TRIM("column") FROM "schema"."table") AS LIMITED_SELECT LIMIT 64                                         
         INFO     [Redshift] Schema = {'column': String_VaryingAlphanum(_notes=[], collation=None)}                                                 schema.py:44
         DEBUG    Running SQL (MsSQL): ('dbo', 'TABLE')                                                                                                base.py:980
                  SELECT TRIM([COLUMN]) FROM [dbo].[TABLE] ORDER BY 1 OFFSET 0 ROWS FETCH NEXT 64 ROWS ONLY                                         
         ERROR    ('42000', "[42000] [Microsoft][ODBC Driver 18 for SQL Server][SQL Server]'TRIM' is not a recognized built-in function name. (195)            __main__.py:344
                  (SQLExecDirectW)")                                                                                                                                          
Traceback (most recent call last):
  File "/home/ubuntu/ingestion/venv/lib/python3.10/site-packages/data_diff/databases/mssql.py", line 227, in _query_cursor
    return super()._query_cursor(c, sql_code)
  File "/home/ubuntu/ingestion/venv/lib/python3.10/site-packages/data_diff/databases/base.py", line 1173, in _query_cursor
    c.execute(sql_code)
pyodbc.ProgrammingError: ('42000', "[42000] [Microsoft][ODBC Driver 18 for SQL Server][SQL Server]'TRIM' is not a recognized built-in function name. (195) (SQLExecDirectW)")

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/ubuntu/ingestion/venv/bin/data-diff", line 8, in <module>
    sys.exit(main())
[...]

PR #886, opened 2024-04-16
update `TRIM` to `LTRIM(RTRIM(...))` for more universal query syntax

MiConnell:
SQL Server versions before 2017 don't support trim, only ltrim and rtrim. this should be universal across databases. 
referenced in #885


glebmezh:
Hi @MiConnell ,

I'm sorry for the delay in following up on this. Thank you for taking the time to open this PR!

We [made a hard decision](https://www.datafold.com/blog/sunsetting-open-source-data-diff) to sunset the data-diff package and won't provide further development or support.

SQL Server 2019+ is now fully supported in Datafold Cloud. As for the earlier SQL Server versions, the lack of UTF8 support  makes efficient diffing very hard.

-Gleb

PR #894, opened 2024-05-14
Update poetry version in Dockerfile to match what we have in poetry.lock

bpaquet:


glebmezh:
Hi @bpaquet,

I'm sorry for the delay in following up on this. Thank you for taking the time to open this PR!

We [made a hard decision](https://www.datafold.com/blog/sunsetting-open-source-data-diff) to sunset the data-diff package and won't provide further development or support.

-Gleb

